PR #263, opened 2026-03-02
docs/test: align processor docs and rendered-output coverage

bdarcus:
This PR tightens processor documentation and test coverage in the areas that were most likely to drift out of sync with rendered behavior.

### Documentation
- Corrected `Disambiguator` docs so year-suffix ordering matches the implementation:
  resolved per-group sort when present, otherwise title-based fallback order.
- Corrected numeric integral rendering docs to describe the real fallback:
  author plus citation number, not author-year.
- Fixed the `calculate_hints` example so it reflects the actual map shape and keys.

### Tests
- Strengthened the new disambiguation coverage to assert rendered citation output,
  not just internal `ProcHints`.
- Added end-to-end checks for:
  - group-aware year-suffix ordering
  - given-name disambiguation in rendered citations
- Removed the stray debug print from the anonymous-work bibliography sort test.
- Kept the anonymous title-sort regression test enabled and passing.

## Why

The goal of this PR is higher confidence that:
- extracted `///` docs are accurate and useful
- passing tests mean rendered output is unlikely to surprise us

The main review feedback was that some new docs overstated or misstated behavior,
and some new tests stopped at internal state instead of asserting user-visible output.
This update closes those gaps.

## Verification

- `cargo fmt`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo nextest run`

### DIFF
--- a/crates/citum-engine/src/processor/disambiguation.rs
+++ b/crates/citum-engine/src/processor/disambiguation.rs
@@ -529,11 +529,16 @@ impl<'a> Disambiguator<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::Processor;
+    use citum_schema::citation::Citation;
     use citum_schema::grouping::{GroupSort, GroupSortKey, SortKey};
+    use citum_schema::options::{Config, ContributorConfig, DisplayAsSort};
     use citum_schema::reference::{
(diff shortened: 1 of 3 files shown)


bdarcus:
@copilot - review this PR

PR #508, opened 2026-04-11
ci: fix Rust dependency caching target

bdarcus:
Configure rust-cache to look in /tmp/citum-target, matching the CARGO_TARGET_DIR used across workflows. This fixes a persistent cache-miss issue that forced full recompilations on every run.


bdarcus:
@copilot can you figure out why this effort to get the CI Rust workflow to properly cache is not working, and fix it? It's regularly taking ~10 minutes, and seems not to be using the cache.

PR #514, opened 2026-04-13
fix(engine): personal communication rendering

bdarcus:
## Summary

- Removes hardcoded personal-communication rendering from the engine (two special-cases in `contributor/mod.rs` and `date.rs`)
- Adds `GeneralTerm::PersonalCommunication` to the schema so styles can reference `term: personal-communication` in templates
- Restructures `apa-7th.yaml` so the "personal communication" label appears correctly:
  - **Integral**: inside the parenthetical group — `Oglethorpe (personal communication, 1733)`
  - **Non-integral**: as a separate term component — `(J. Oglethorpe, personal communication, 1733)`
- Fixes `examples/chicago-note-converted.yaml`: the Oglethorpe record previously embedded author and recipient in a `title` field (Zotero export artifact); now uses proper `contributors` with `author` + `recipient` roles
- Creates `docs/specs/PERSONAL_COMMUNICATION_CITATION.md` covering APA 7th, Chicago, and MLA formats

## Test plan

- [x] `cargo nextest run` — 1020 tests pass
- [x] `cargo fmt --check && cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] Verified render output against APA 7th expected format

## Design note

The previous engine special-cases violated the "style declares behavior, processor stays dumb" principle. All rendering logic now lives in the style YAML template. The `personal-communication` locale term is registered in `GeneralTerm` and `Terms::en_us()` so it works even when the on-disk locale file fails to parse (pre-existing issue, separate from this PR).

### DIFF
--- a/crates/citum-schema-style/src/locale/mod.rs
+++ b/crates/citum-schema-style/src/locale/mod.rs
@@ -814,6 +814,7 @@ impl Locale {
             GeneralTerm::Edition => "edition",
             GeneralTerm::Section => "section",
             GeneralTerm::OriginalWorkPublished => "original-work-published",
+            GeneralTerm::PersonalCommunication => "personal-communication",
         }
     }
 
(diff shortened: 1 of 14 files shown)


crates/citum-schema-style/src/locale/mod.rs
```diff
@@ -1340,6 +1341,9 @@ impl Locale {
             "deposited" => Some(GeneralTerm::Deposited),
             "review-of" | "review_of" | "review of" => Some(GeneralTerm::ReviewOf),
             "original-work-published" => Some(GeneralTerm::OriginalWorkPublished),
+            "personal-communication" | "personal_communication" | "personal communication" => {
```
bdarcus: Why are we allowing different string representations here and elsewhere in the file? I thought we control this to some level? Even if must keep it for some reason, pros and cons of moving this logic to a simple normalization helper?

crates/citum-schema-style/src/locale/types.rs
```diff
@@ -179,13 +181,22 @@ impl Terms {
...
-                    long: "no date".into(),
-                    short: "n.d.".into(),
-                },
-            )]),
+            general: std::collections::HashMap::from([
+                (
+                    GeneralTerm::NoDate,
+                    SimpleTerm {
+                        long: "no date".into(),
+                        short: "n.d.".into(),
```
bdarcus, line 189: Why are we using locale-specific strings here?

crates/citum-schema-style/src/locale/types.rs
```diff
@@ -179,13 +181,22 @@ impl Terms {
...
+                    },
+                ),
+                (
+                    GeneralTerm::PersonalCommunication,
+                    SimpleTerm {
+                        long: "personal communication".into(),
+                        short: "pers. comm.".into(),
```
bdarcus, line 196: Why are we using locale-specific strings here?

PR #530, opened 2026-04-17
feat(spec): original publication relation support

bdarcus:
Draft a specification to expand the original WorkRelation field to all InputReference variants. This ensures original-date and other original publication metadata are correctly captured and rendered across all types, resolving gaps identified in Chicago 18th benchmarks.


bdarcus (commented):
Addressed Copilot feedback by reformatting the spec to match the template, adding missing sections, and correcting the memory usage risk assessment. Amending the commit now.

PR #544, opened 2026-04-19
feat(locale): add gender-aware term resolution

bdarcus:
## Summary
- add grammatical gender support for locale terms, noun metadata, template overrides, and contributor reference data
- make mixed-gender contributor groups prefer neutral/common role labels instead of falling back to masculine-specific forms
- add Spanish-focused contributor-role coverage, keep Arabic ordinal coverage, and regenerate schemas

## Testing
- cargo fmt --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo nextest run
- cargo run --bin citum --features schema -- schema --out-dir docs/schemas

Refs: csl26-y3kj


bdarcus:
A few things:

1. I don't see a spanish locale file, which we should add.
2. There's no doc updates, including on the public docs site (e.g. examples, style guide).

PR #615, opened 2026-05-04
docs(spec): interactive server API and bean

bdarcus:
The current `render_citation` API processes citations one at a time with no knowledge of document context. That makes ibid, back-references, and note-number inference impossible. This spec defines the two-tier API that fixes that.

## What this spec covers

**Tier 1 — `format_document`** is the primary interface. Modelled on Pandoc/citeproc: one call takes the full ordered citation list plus references, returns all formatted citations and the bibliography in one shot. Stateless, works in stdio, HTTP, and WASM. This is what most callers should use.

**Tier 2 — Session lifecycle** is for word processors that can't afford to re-send a large reference library on every keystroke. `open_session` → `put_references` / `insert_citations_batch` → per-edit mutations → `close_session`. Stateful, HTTP-only, behind a `session` feature flag.

Both tiers share types in `citum-engine`; `citum-bindings` and `citum-server` stay as thin adapters.

## Key design decisions worth reviewing

- **`StyleInput` union** (`path | embedded | yaml`) replaces the bare `style` string, so WASM callers can pass inline YAML and server callers can use registry IDs without adapter-specific leakage. A `url` variant is reserved for the distributed resolver work in csl26-r8d2.
- **`CitationOccurrence` mirrors the existing `Citation` schema** — including `mode: integral | non-integral` for narrative citations, `CitationLocator` for structured locators, and `integral_name_state` per-item. No impedance mismatch at the engine boundary.
- **Session results use a unified envelope** with `version`, `affected_citations` (changed-only), `renumbering_occurred`, and structured `warnings`. Clients only re-render what changed.
- **`DocumentOptions`** covers per-document overrides: bibliography grouping (reusing `BibliographyGroup` from the existing engine), annotation maps (`annotations: Record<refId, text>` + `annotation_format`), and Pandoc-equivalent flags. The Pandoc flags are marked draft — whether to adopt them is an open question.

## What is *not* in this spec

- Implementation — this is a design document only
- Streaming progress (Tier 3) — deferred; would use SSE
- `validate_document` — useful future work, not scoped here
- Changes to `render_citation`, `render_bibliography`, or `validate_style` — those stay as-is

## Open questions for reviewers

1. **`DocumentOptions` Pandoc fields** — `suppress_bibliography`, `link_citations`, `link_bibliography`, `notes_after_punctuation` are flagged as TBD. Should Citum adopt any of these, and if so, do they belong in `document_options` or at the top level?
2. **`annotated_bibliography` mode** — the engine currently takes an explicit annotation map; is there a case for also inferring annotations from the reference's `annote` field, or should the caller always supply them?
3. **Session TTL** — 30-minute default is a placeholder. What's the right default for a word processor context?

### DIFF
--- a/.beans/csl26-isrv--server-interactive-api.md
+++ b/.beans/csl26-isrv--server-interactive-api.md
@@ -1,0 +1,34 @@
+---
+# csl26-isrv
+title: Interactive server API — document-batch and session modes
+status: todo
+type: feature
+priority: normal
+created_at: 2026-05-04T00:00:00Z
+updated_at: 2026-05-04T12:00:04Z
+---
+
+Implement the interactive server API as specified in
+`docs/specs/SERVER_INTERACTIVE_API.md`.
+
+Two tiers:
+
+1. **`format_document`** (stateless, document-shaped) — primary delivery.
+   A single call takes the full ordered citation list and reference set,
+   returns all formatted citations plus bibliography. Enables correct
+   note-position inference, ibid, and disambiguation — impossible with the
+   current per-citation `render_citation` method.
+
+2. **Session lifecycle** (`open_session` … `close_session`) — advanced,
+   behind a `session` feature flag (implies `http`). Amortizes style
+   parsing and deserialization for word processors with large bibliographies.
+
+Core types (`FormatDocumentRequest`, `DocumentSession`, etc.) go in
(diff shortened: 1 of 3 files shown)


bdarcus:
To resolve before finalizing and implementing:

  - csl26-r8d2
  - csl26-extg

PR #715, opened 2026-05-15
docs(spec): forward-compat contract + tests

bdarcus:
## Summary

Confirms (with a measurement, not a claim) what an older Citum engine actually does today when it reads a newer style/reference/locale, lands the canonical forward-compatibility contract in `docs/specs/FORWARD_COMPATIBILITY.md`, and files five follow-up beans — one per snapshot gap.

No engine behavior changes here. The test passes against the snapshot of current state. Follow-up PRs close gaps by editing the snapshot, which is the handoff artifact.

## Why now

Crates are about to be published and will continue rapid iteration. The desired contract — "new features should warn rather than break; only template grammar changes may hard-fail" — was claimed across DESIGN_PRINCIPLES, SCHEMA_VERSIONING, ENUM_VOCABULARY_POLICY, and EXTENSIBILITY_STRATEGY, but the implementation contradicts it in several places. This PR makes the contract single-sourced, the behavior measurable, and the gaps trackable.

## Snapshot at a glance

`crates/citum-engine/tests/snapshots/forward_compat_gaps.snap` — 13 rows, 7 gaps, 6 controls aligned.

| Class of row | Count | Notes |
|---|---|---|
| `declared=SoftDegrade observed=HardFail` | 6 | Attribute enums (1, 2, 3, 4), new option key (5), new top-level section (6), new locale term key (8). Fix path: tolerant deserializer + capture-unknown-fields wrapper. |
| `declared=SoftDegrade observed=Pass` (silent drop) | 1 | New reference field (7). serde+`#[serde(tag)]` swallows unknown keys with no warning. |
| `declared=HardFail observed=HardFail` (correct opt-out) | 3 | Unknown `InputReference::class` (2b), new TemplateComponent variant (11), malformed template shape (12). |
| `declared=Pass observed=Pass` (correct control) | 2 | Namespaced `custom.*` (9), `version: "99.0"` on otherwise-valid style (10). |

## Type-system constraint worth flagging

`InputReference` uses `#[serde(tag = "class")]` at one place. The discriminator cannot soft-degrade by ordinary tolerant-enum tricks because there is no struct shape to deserialize into for an unknown class. The spec calls this out explicitly and presents two options (catch-all `Unknown` variant, recommended; or major-bump category). Tracked separately in **csl26-1bdr**.

Every other enum the prior reviews tightened (`MonographType`, `SerialComponentType`, `ContributorRole`, `TermForm`, …) is an *attribute* inside an already-dispatched struct and is **not** a dispatch constraint — those can carry an `Unknown(String)` fallback safely.

## Follow-up beans

| Bean | Snapshot rows owned | Description |
|---|---|---|
| `csl26-ld6e` | 01, 02, 03, 04 | Tolerant deserializer for attribute enums |
| `csl26-0ksu` | 05, 06 | Capture-unknown-fields wrapper for style schemas |
| `csl26-acfh` | 07 | Surface silent unknown-field acceptance on reference data |
| `csl26-o1z5` | 08 | Tolerant locale term-key lookup |
| `csl26-1bdr` | 02b | `InputReference` unknown-class escape hatch (design + impl) |

Each bean is `blocked_by: csl26-2a0b`. When an impl PR lands, it deletes the row from the snapshot — the diff is the proof.

## Test plan

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cargo nextest run` — 1270 / 1270 pass
- [x] `cargo nextest run -p citum-engine --test forward_compatibility` — 1 / 1 pass
- [ ] Reviewer reads `docs/specs/FORWARD_COMPATIBILITY.md` and confirms the contract matches intent
- [ ] Reviewer scans the snapshot and confirms each `GAP` row has a sensible follow-up bean
- [ ] Reviewer confirms the four cross-link edits read cleanly in context

## Snapshot update workflow (for future PRs)

If an impl PR intentionally changes a row's `observed` outcome:

```bash
UPDATE_FORWARD_COMPAT_SNAPSHOT=1 cargo nextest run -p citum-engine --test forward_compatibility
git add crates/citum-engine/tests/snapshots/forward_compat_gaps.snap
```

Drift without explicit update fails CI.

### DIFF
--- a/docs/specs/FORWARD_COMPATIBILITY.md
+++ b/docs/specs/FORWARD_COMPATIBILITY.md
@@ -97,31 +97,36 @@ follow-up implementation beans; this spec only fixes the contract.
 ## Scope table
 
 Each row is anchored to one or more cases in
-`crates/citum-engine/tests/forward_compatibility.rs`. The "Currently
-observed" column reflects what the engine does **today**, not what the
-contract demands.
+`crates/citum-engine/tests/forward_compatibility.rs`. The columns below
+measure **loader behavior** and align row-for-row with
+`crates/citum-engine/tests/snapshots/forward_compat_gaps.snap` — the
+truth-of-record. End-to-end user-visible outcomes may add a warning via
+`citum check` on top of a loader `Pass`; see row 10.
 
-| # | Category | Example | Desired | Currently observed | Follow-up |
+| # | Category | Example | Declared | Observed | Follow-up |
 |---|---|---|---|---|---|
-| 1 | Attribute enum in template | `contributor: producer` (new `ContributorRole`) | `SoftDegrade` | `HardFail` (raw serde error) | Tolerant enum deserializer |
-| 2 | Attribute enum in data | `class: monograph, type: dance-performance` | `SoftDegrade` | `HardFail` | Tolerant enum deserializer |
-| 2b | Top-level `class` value | `class: dance-performance` | `HardFail` (opt-out, major bump) **or** `SoftDegrade` via catch-all variant | `HardFail` | InputReference discriminator design |
-| 3 | Locale form | `form: vocative` | `SoftDegrade` (fall back to `Long`) | `HardFail` | Tolerant enum deserializer |
-| 4 | Date form | `form: month-and-day` (new `DateForm`) | `SoftDegrade` | `HardFail` | Tolerant enum deserializer |
-| 5 | New style option key | `options.contributors.future-key: true` | `SoftDegrade` (ignore field, warn) | `HardFail` (`deny_unknown_fields`) | Capture-unknown-fields wrapper |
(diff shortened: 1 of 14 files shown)


docs/specs/FORWARD_COMPATIBILITY.md
```diff
@@ -0,0 +1,278 @@
...
+serde_json::Map<String, Value> })`. A brand-new class round-trips its
+data, emits a `SoftDegrade` warning, and degrades to a generic
+rendering path (likely "render title + author + year, drop everything
+else"). Preserves the spec's headline rule that new features warn
+rather than break.
+
+**Option B — Major-bump category.** Declare new top-level classes as a
+second opt-out category alongside template grammar. Producers must bump
+major and accept that older engines hard-fail. Simpler to implement;
+worse user experience.
```
bdarcus: At this point, there is no concern for backward-compatibility, and therefore breakage.  The priority is getting this right going forward.

Does that change this assessment at all?

PR #771, opened 2026-05-21
docs(spec): style editions and families strategy

bdarcus:
## Summary

New spec at [`docs/specs/STYLE_EDITIONS_AND_FAMILIES.md`](docs/specs/STYLE_EDITIONS_AND_FAMILIES.md) codifying long-term policy for how Citum handles style editions and families — the questions CSL 1.0 never had to answer because it lacks explicit inheritance.

Answers five questions the existing architecture docs left implicit:

- **§1 Edition inheritance shape.** Editions are siblings under an optional hidden archetype; no edition-on-edition stacking (fragile-base-class risk; cf. MLA 7→8 container shift).
- **§2 Embedding retention.** Current + immediately prior edition (N + N−1) compiled in; older editions move to `styles/legacy-editions/` or are removed under a major schema bump.
- **§3 Retirement workflow.** Single-PR coordinated rewrites with a CI integrity check forbidding `extends:` of non-existent bases. `deprecated:` marker + SoftDegrade warnings during the deprecation window.
- **§4 Version field.** Keep `info.edition`, `info.citum_version`, and the `StyleBase` key. Reject a new `info.version` field — it would duplicate signals git history and existing fields already carry.
- **§5 Tooling impact.** `citum-migrate` edition detection and `/style-*` skill changes documented but deferred to follow-up beans in §7.

Pure docs PR; no Rust, schema, or test surface touched.

Cross-references added back from [`STYLE_PRESET_ARCHITECTURE.md`](docs/specs/STYLE_PRESET_ARCHITECTURE.md), [`STYLE_TAXONOMY.md`](docs/specs/STYLE_TAXONOMY.md), [`FORWARD_COMPATIBILITY.md`](docs/specs/FORWARD_COMPATIBILITY.md).

Bean: `csl26-atyp` (completed, archived in this commit).

## Test plan

- [x] Spec renders correctly on GitHub.
- [x] All four cross-references resolve.
- [x] No CI surface beyond markdown lint (no Rust changed).


docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+
+The spec is normative for new editions, new family heads, and any future
+retirement of a `StyleBase`. Mechanism changes that are not needed immediately
+are tracked as follow-up beans (see §7).
+
+---
+
+## §1 Edition inheritance shape
+
+**Recommendation.** Editions are siblings under an optional hidden archetype,
```
bdarcus, line 36: What does "hidden" mean here? You mean they are styles tagged in some way in the registry where they aren't listed in the style list?

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+not stacked edition-on-edition.
+
+### Rule
+
+When a publisher releases a new edition of a style family (e.g. APA 7 → APA 8),
+the new edition is added as a new Tier-1 `StyleBase` variant with its own
+YAML in `styles/embedded/`. It **must not** declare `extends: <prior-edition>`.
```
bdarcus, line 59: Why? Aren't new styles always expressed as delta from previous styles?

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+introduced (e.g. `ApaAuthorDateArchetypeCore`). Both editions then
+`extends: apa-author-date-archetype-core`. Archetypes follow the existing
+`*-Core` pattern from publisher families (`ElsevierHarvardCore`,
+`SpringerBasicAuthorDateCore`, etc.) and are never user-selectable in the
+registry.
+
+### Rationale
+
+- **Fragile-base-class risk.** A fix to APA 7's bibliography template would
+  silently re-tune APA 8 under edition-on-edition stacking. The two editions
```
bdarcus, line 105: This addresses my question above. So neither APA 6 or 7 now should include an `extends`, unless we define a common "hidden" archetype style which both `extends:`?

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+### Example: APA 8 in 2030
+
+| File | `extends:` | Notes |
+|---|---|---|
+| `styles/embedded/apa-7th.yaml` | none | retained as the prior-edition base per §2 |
+| `styles/embedded/apa-8th.yaml` | none | new Tier-1 base; full structure |
+| `styles/embedded/apa-author-date-archetype-core.yaml` | none | *optional* hidden archetype, only if both editions converge enough to justify it |
+| `styles/journal-foo.yaml` | `apa-7th` | wrapper stays on APA 7 until its maintainer bumps it |
+
+External reference: Perplexity discussion at the bottom of bean `csl26-atyp`
```
bdarcus: The Perplexity notes are not included in the bean; recast this, and if necessary, reference it's sources.

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
bdarcus: Defined explicitly in §1 of 6553895: a *hidden* base is in `StyleBase` (reachable via `extends:`) but not enumerated in `registry/default.yaml`, so it doesn't appear in `citum styles` / `citum registry list` / the citum-hub picker. Cross-referenced `STYLE_REGISTRY.md` for the discovery surface.

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
bdarcus: Good question — clarified in 6553895. The publisher's release-notes *delta* is an authoring input that gets consumed once when writing the new edition's YAML; it isn't encoded as a live `extends:` edge. Doing the latter would bind APA 8 to whatever APA 7's structure happens to be at any moment, even after the publisher has moved on. Added a paragraph after the Recommendation: 'The authoring input is a delta; the runtime representation is a peer.'

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
bdarcus: Yes, exactly that reading. Made it explicit in 6553895 by restructuring the §1 example into two clearly-labeled cases — Case A (both editions standalone, no `extends:`) and Case B (both editions `extends:` a hidden archetype). The rule 'edition N+1 never `extends:` edition N' is now stated in both cases.

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
bdarcus: Fixed in 6553895 — dropped the bean reference entirely and now cite the two public sources (Sonatype hierarchy best practices, Wikipedia composition-over-inheritance) directly with a one-sentence summary of what each contributes to the argument.

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+---
+
+## §5 Tooling impact
+
+The spec does not require any tooling change to be Active. The two changes
+below are recommended and tracked as follow-up beans in §7.
+
+### `citum-migrate`
+
+Today edition detection is manual: the `extract_info` helper in
```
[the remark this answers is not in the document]
bdarcus: Fixed in 6553895 — renamed to `InfoExtractor::extract` (verified at `crates/citum-migrate/src/info_extractor.rs:14`).

docs/specs/STYLE_EDITIONS_AND_FAMILIES.md
```diff
@@ -0,0 +1,313 @@
...
+**Recommendation.** Keep the fields that exist. Do not introduce a new style
+version.
+
+### Rule
+
+The structured edition identity is the `StyleBase` *key*
+(`apa-7th`, `chicago-notes-18th`, …). The display surface is `info.edition`
+("7th", "18th edition"), free text by design. Engine compatibility is
+declared by `info.citum_version` as a `semver::VersionReq` (see
+`citum_version` in `crates/citum-schema-style/src/style/metadata.rs`).
```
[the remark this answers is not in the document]
bdarcus: Fixed in 6553895 — reworded to 'an optional `semver::VersionReq`-compatible string', matching the `Option<String>` field declaration and doc comment at `crates/citum-schema-style/src/style/metadata.rs:159`.

PR #817, opened 2026-05-26
feat: per-document config overrides

bdarcus:
## Summary

Implements the per-document configuration override spec
(`docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md`) via three commits:

- **Structs + frontmatter parsing**: `DocumentOptionsOverride`,
  `DocumentBibliographyOverride`, `DocumentSortPartitioningOverride`
  (all-optional sparse overlay). Added to `DocumentFrontmatter` for
  Djot; Markdown parser now shares `parse_frontmatter` from djot.
  Frontmatter parse errors emit `eprintln!` warning instead of
  silently discarding.
- **Pipeline wiring**: `process_document()` applies `frontmatter_options`
  — locale, integral-name-memory, and bibliography overrides (sort
  partitioning, repeated-author rendering, label mode/wrap). Added
  `Style::apply_scoped_options()` bridge method so the engine crate can
  trigger the schema-crate scoped-options pass.
- **CLI `--locale` / `-L` flag**: `render doc` accepts an explicit
  locale that overrides `options.locale` from frontmatter.

Demo updated: `docs/demo.djot` now carries an `options.integral-name-memory`
block; `docs/demo.html` reflects the live engine output with short-form
name repetition for Chen (2017/2020). Bibliography sections now use
`class="citum-bibliography"` directly on the `<section>` elements so
headings are properly nested inside the styled section (fixes a visual
separation issue where the CSS `border-top` was rendering between the
heading and its entries).

## Test plan

- [x] 8 new unit tests in `types.rs` cover struct deserialization,
  `apply_to` field application, and locale/integral-name-memory
  precedence rules
- [x] 4 unit tests in `render/mod.rs` for `extract_frontmatter_locale`
- [x] All 1415 existing tests pass
- [x] `cargo fmt --check` + `cargo clippy -D warnings` clean
- [x] `docs/demo.html` regenerated from `citum render doc` with
  name-memory enabled; bibliography section nesting fixed

### DIFF
--- a/crates/citum-cli/src/commands/render/mod.rs
+++ b/crates/citum-cli/src/commands/render/mod.rs
@@ -53,15 +53,25 @@ pub(super) fn run_render_doc(args: RenderDocArgs) -> CliResult {
         );
     }
 
+    // Read document first so we can extract frontmatter locale for processor setup.
+    let doc_content = fs::read_to_string(&args.input)?;
+
+    // CLI --locale takes precedence over frontmatter options.locale.
+    // Avoid cloning args.locale: only call extract_frontmatter_locale when needed.
+    let fm_locale = args
+        .locale
+        .is_none()
+        .then(|| extract_frontmatter_locale(&doc_content))
+        .flatten();
+    let effective_locale = args.locale.as_deref().or(fm_locale.as_deref());
+
     let processor = create_processor(
         style_obj,
         bibliography,
         &args.style,
         args.no_semantics,
-        None,
+        effective_locale,
     )?;
-
-    let doc_content = fs::read_to_string(&args.input)?;
     let output = match args.input_format {
         InputFormat::Djot => render_doc_with_output_format(
             &processor,
@@ -307,6 +317,44 @@ fn render_refs_json(
     }
 }
 
+/// Extract the `options.locale` field from document YAML frontmatter, if present.
+///
+/// Returns `None` when the document has no frontmatter, the frontmatter has no
+/// `options:` block, or `options.locale` is absent. Parsing errors are silently
+/// ignored — the engine will fall back to the style's default locale.
+fn extract_frontmatter_locale(content: &str) -> Option<String> {
+    use serde::Deserialize;
+
+    #[derive(Deserialize)]
+    struct FrontmatterOptions {
+        locale: Option<String>,
+    }
+
+    #[derive(Deserialize)]
+    struct MinimalFrontmatter {
+        options: Option<FrontmatterOptions>,
+    }
+
+    let trimmed = content.trim_start();
+    if !trimmed.starts_with("---") {
+        return None;
+    }
+    #[allow(clippy::string_slice, reason = "'---' is 1-byte ASCII")]
+    let after_opening = &trimmed[3..];
+    // Use "\n---" so embedded "---" in YAML values (e.g. abstract: "foo---bar")
+    // does not falsely close the frontmatter block.
+    let closing_pos = after_opening.find("\n---").map(|p| p + 1)?;
+    #[allow(
+        clippy::string_slice,
+        reason = "closing_pos is '\\n' offset + 1, which lands on ASCII '-'"
+    )]
+    let frontmatter_str = &after_opening[..closing_pos];
+    serde_yaml::from_str::<MinimalFrontmatter>(frontmatter_str)
+        .ok()?
+        .options?
+        .locale
+}
+
 #[cfg(test)]
 #[allow(
     clippy::unwrap_used,
@@ -547,6 +595,44 @@ grammar-options:
         assert_eq!(
             parsed.legacy_term_aliases.get("page").map(String::as_str),
             Some("term.page-label")
+        );
+    }
+
+    #[test]
+    fn test_extract_frontmatter_locale_present() {
+        let content = "---\noptions:\n  locale: de-DE\n---\nBody text.";
+        assert_eq!(
+            extract_frontmatter_locale(content),
+            Some(String::from("de-DE"))
+        );
+    }
+
+    #[test]
+    fn test_extract_frontmatter_locale_absent_options() {
+        let content = "---\ntitle: My Doc\n---\nBody text.";
+        assert!(extract_frontmatter_locale(content).is_none());
+    }
+
+    #[test]
+    fn test_extract_frontmatter_locale_no_frontmatter() {
+        let content = "No frontmatter here.";
+        assert!(extract_frontmatter_locale(content).is_none());
+    }
+
+    #[test]
+    fn test_extract_frontmatter_locale_options_without_locale() {
+        let content = "---\noptions:\n  bibliography:\n    label-mode: numeric\n---\nBody.";
+        assert!(extract_frontmatter_locale(content).is_none());
+    }
+
+    #[test]
+    fn test_extract_frontmatter_locale_embedded_dashes_in_value() {
+        // Regression: inline "---" inside a YAML value must not be treated as
+        // the closing fence. The correct closing "---" is on its own line.
+        let content = "---\ntitle: \"pre---post\"\noptions:\n  locale: fr-FR\n---\nBody.";
+        assert_eq!(
+            extract_frontmatter_locale(content),
+            Some(String::from("fr-FR"))
         );
     }
 }
--- a/crates/citum-engine/src/processor/document/djot/mod.rs
+++ b/crates/citum-engine/src/processor/document/djot/mod.rs
@@ -23,8 +23,12 @@ pub struct DjotParser;
 impl CitationParser for DjotParser {
     fn parse_document(&self, content: &str, locale: &Locale) -> ParsedDocument {
         // Try to parse frontmatter and get remaining content
-        let (frontmatter, remaining_content) = parse_frontmatter(content);
+        let (frontmatter_result, remaining_content) = parse_frontmatter(content);
         let body_start = content.len() - remaining_content.len();
+        let (frontmatter, frontmatter_error) = match frontmatter_result {
+            Ok(fm) => (fm, None),
+            Err(e) => (None, Some(e)),
+        };
 
         let (manual_note_references, manual_note_labels, footnote_definitions) =
             scan_manual_notes(remaining_content);
@@ -53,12 +57,25 @@ impl CitationParser for DjotParser {
         let bibliography_blocks = scan_bibliography_blocks(remaining_content);
 
         let frontmatter_groups = frontmatter.as_ref().and_then(|fm| fm.bibliography.clone());
+        let frontmatter_options = frontmatter.as_ref().and_then(|fm| fm.options.clone());
+        // Legacy top-level fields are superseded by their `options.*` counterparts.
         let frontmatter_integral_name_memory = frontmatter
             .as_ref()
-            .and_then(|fm| fm.integral_name_memory.clone());
+            .and_then(|fm| fm.integral_name_memory.clone())
+            .filter(|_| {
+                frontmatter_options
+                    .as_ref()
+                    .and_then(|o| o.integral_name_memory.as_ref())
+                    .is_none()
+            });
         let frontmatter_org_abbreviation_memory = frontmatter
-            .as_ref()
-            .and_then(|fm| fm.org_abbreviation_memory.clone());
+            .and_then(|fm| fm.org_abbreviation_memory)
+            .filter(|_| {
+                frontmatter_options
+                    .as_ref()
+                    .and_then(|o| o.org_abbreviation_memory.as_ref())
+                    .is_none()
+            });
         ParsedDocument {
             citations,
             manual_note_order,
@@ -68,6 +85,8 @@ impl CitationParser for DjotParser {
             frontmatter_groups,
             frontmatter_integral_name_memory,
             frontmatter_org_abbreviation_memory,
+            frontmatter_options,
+            frontmatter_error,
             body_start,
         }
     }
--- a/crates/citum-engine/src/processor/document/pipeline.rs
+++ b/crates/citum-engine/src/processor/document/pipeline.rs
@@ -6,10 +6,10 @@ SPDX-FileCopyrightText: © 2023-2026 Bruce D'Arcus and Citum contributors
 //! High-level document-processing orchestration.
 
 use super::output::{
-    HtmlPlaceholderRegistry, RenderedDocumentBody, append_document_bibliography,
-    bibliography_block_placeholder, render_document_bibliography_block_replacement,
-    rewrite_document_markup_for_typst, rewrite_group_headings_for_document,
-    stage_document_bibliography_blocks,
+    append_document_bibliography, bibliography_block_placeholder,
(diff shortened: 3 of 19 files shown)


docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+syntax for plain-text documents, and the equivalent surface for GUI consumers
+such as word-processor plugins.
+
+## Scope
+
+**In scope:**
+
+- Defining the eligible option set and the reasoning behind each inclusion
+- YAML frontmatter syntax (`options:` block)
+- GUI consumer contract (equivalent logical shape, wire format deferred)
```
bdarcus: Wouldn't it make sense to use serde here as we do elsewhere, so the frontmatter is defined one in Rust code, and serde supports YAML, JSON, and CBOR? So the contract would not be deferred; it'd be the same in doc as YAML as it would external as JSON or CBOR?

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+both by authors preparing manuscripts (where the dash should be suppressed)
+and by publishers rendering final output (where the dash is appropriate),
+this is a clear case for per-document control.
+
+Values: `full` (always render the full contributor list), `dash`, `dash-with-space`.
+
+### `bibliography.label-mode`
+
+The numeric vs. author-date vs. none label distinction is a document-class
+concern. A researcher may use the same CSL-derived style to produce a journal
```
bdarcus: CSL? What does that have to do with this project?

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+feature. Vancouver uses square brackets; ACS uses superscripts; NLM uses
+no wrap. Overriding `label-wrap` per document can produce output that
+superficially resembles a different named style without declaring it as such,
+which may mislead readers about the citation convention in use.
+
+This does not disqualify the option — bracket style is still presentation-layer
+— but authors should use it with care and be aware that the resulting output
+may not conform to any named style.
+
+### `bibliography.date-position`
```
bdarcus, line 188: I'm not convinced by this explanation. Move it to a later section for possible future consideration.

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+match the in-text citation key `(Smith 2020)` to the bibliography entry. If
+`date-position` is overridden to `after-title` or `terminal` in a document
+that uses author-date in-text citations, the bibliography will still be
+*correct*, but readers will need to scan further to find the date, degrading
+usability.
+
+Implementation should emit a warning when `date-position` is overridden on a
+style whose `config.processing` uses an author-date citation mode.
+
+### `bibliography.title-terminator`
```
bdarcus, line 198: also not convinced by this

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+Terminal punctuation after bibliography titles (period, comma, none) is
+micro-formatting. It is a safe per-document override and meets all eligibility
+criteria.
+
+Note: when `title-terminator` and `date-position` are both overridden
+simultaneously, the interaction should be tested. A title ending with a period
+before a repositioned date may produce `Title. 2020.` rather than `Title. 2020`
+or `Title 2020.` depending on the rendering order.
+
+### `citation.label-wrap`
```
bdarcus, line 205: also not convinced by this

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+
+Same criteria and caveat as `bibliography.label-wrap`. Citation and bibliography
+label wraps should normally be overridden together to keep in-text and
+list-style labels consistent.
+
+### `citation.group-delimiter`
```
bdarcus, line 211: Are you saying with this there's no need for a vancouver style? Not sure about this one either, but might be convinced.

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+fields inherit the style's defaults.
+
+```yaml
+---
+options:
+  locale-override: de-DE
+  integral-name-memory:
+    enabled: false
+  bibliography:
+    sort-partitioning:
```
bdarcus, line 234: shouldn't this be called "grouping"?

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+        secondary:
+          label: "Secondary Sources"
+    label-mode: numeric
+    label-wrap: brackets
+    date-position: after-author
+    title-terminator: period
+    repeated-author-rendering: full
+  citation:
+    label-wrap: brackets
+    group-delimiter: semicolon
```
bdarcus: Per my comments above, I'm not convinced by any of this in the first pass.

docs/specs/PER_DOCUMENT_CONFIG_OVERRIDES.md
```diff
@@ -0,0 +1,303 @@
...
+`integral-name-memory` remains at the top level of `options:` (not nested
+under `citation:` or `bibliography:`) for backward compatibility with existing
+documents.
+
+## GUI Access Surface
+
+Word-processor plugins and other GUI consumers send the same logical options
+as the YAML frontmatter, encoded as structured data (typically a JSON object
+with the same shape as the YAML). The wire format for transmitting these
+overrides to the engine is defined in `docs/specs/SERVER_INTERACTIVE_API.md`.
```
bdarcus: see first comment

crates/citum-engine/src/processor/document/pipeline.rs
```diff
@@ -36,14 +36,44 @@ impl Processor {
...
+        let effective_integral_override = parsed
+            .frontmatter_options
+            .as_ref()
+            .and_then(|o| o.integral_name_memory.as_ref())
+            .or(parsed.frontmatter_integral_name_memory.as_ref());
+        let owned_integral =
+            self.processor_with_document_integral_name_override(effective_integral_override);
+
+        // Apply org-abbreviation-memory override from the options block.
+        let effective_org_override = parsed
```
bdarcus, line 55: **Bug: `org-abbreviation-memory:` at top level is now silently dropped**

`integral-name-memory:` retains a legacy compat path (`frontmatter_integral_name_memory` field on `ParsedDocument` + the `.or()` at line 45). Org-abbreviation gets neither: `DocumentFrontmatter` no longer has an `org_abbreviation_memory` field, and the pipeline only reads it from `frontmatter_options.org_abbreviation_memory`.

Documents that previously used:
```yaml
---
org-abbreviation-memory:
  enabled: true
  scope: chapter
---
```
will silently lose that override (serde skips unknown fields on `DocumentFrontmatter` — no `deny_unknown_fields`). Consider adding a legacy `frontmatter_org_abbreviation_memory` field mirroring the integral-name pattern, or emitting a migration warning.

crates/citum-engine/src/processor/document/types.rs
```diff
@@ -118,61 +123,213 @@ impl DocumentOrgAbbreviationOverride {
...
+#[derive(Debug, Clone, Default, Deserialize, Serialize, PartialEq)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub struct DocumentOptionsOverride {
+    /// BCP 47 locale ID to use as the base locale for this document.
+    ///
+    /// Replaces the style's default locale entirely. Handled in the CLI layer;
+    /// `apply_to` does not touch locale.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub locale: Option<String>,
```
bdarcus, line 268: **Design gap: `locale` only works via CLI; WASM/FFI consumers silently ignore it**

The doc comment acknowledges this (`Handled in the CLI layer; `apply_to` does not touch locale`), but it means `citum-hub` (WASM bridge) and `citum-labs` (FFI) callers who build a `Processor` and call `process_document` directly will silently use the style's default locale even when `options.locale` is set in frontmatter.

The spec promises `DocumentOptionsOverride` is the single contract for all consumers. The current approach requires each consumer surface to independently re-implement `extract_frontmatter_locale`. Flagging for a follow-up: either expose `parse_frontmatter` from the engine's public API or have `process_document` accept an optional locale override that the engine resolves from `frontmatter_options.locale`.

crates/citum-cli/src/commands/render/mod.rs
```diff
@@ -307,6 +317,44 @@ fn render_refs_json(
     }
 }
 
+/// Extract the `options.locale` field from document YAML frontmatter, if present.
+///
+/// Returns `None` when the document has no frontmatter, the frontmatter has no
```
bdarcus, line 322: **Reuse: duplicates `parse_frontmatter` boundary detection + mirrors engine types**

This function re-implements the `---` / `find("---")` frontmatter scanner already in `crates/citum-engine/src/processor/document/djot/parsing.rs:368`, and defines `MinimalFrontmatter` / `FrontmatterOptions` that mirror `DocumentFrontmatter` / `DocumentOptionsOverride` already in the engine.

Two divergence risks:
1. If `parse_frontmatter` gains support for `...` as a closing marker or CRLF normalization, this function won't.
2. If `DocumentOptionsOverride.locale` is renamed or moved into a sub-struct, this function silently returns `None` instead of compiling loudly.

Longer-term fix: expose a thin `parse_document_locale(content: &str) -> Option<String>` from `citum-engine`'s public API, or expose `DocumentOptionsOverride` and `parse_frontmatter` as `pub` so CLI can call `parse_frontmatter(content).0?.options?.locale` directly.

crates/citum-engine/src/processor/document/djot/mod.rs
```diff
@@ -53,12 +53,17 @@ impl CitationParser for DjotParser {
         let bibliography_blocks = scan_bibliography_blocks(remaining_content);
 
         let frontmatter_groups = frontmatter.as_ref().and_then(|fm| fm.bibliography.clone());
+        let frontmatter_options = frontmatter.as_ref().and_then(|fm| fm.options.clone());
+        // Legacy top-level `integral-name-memory:` is superseded by
+        // `options.integral-name-memory` when both are present.
         let frontmatter_integral_name_memory = frontmatter
```
bdarcus, line 62: **Reuse: identical precedence block copy-pasted into `markdown.rs`**

This 8-line block (extracting `frontmatter_options` + suppressing legacy `integral_name_memory` via `.filter()`) is verbatim in both `djot/mod.rs` and `markdown.rs`. A future third parser format would copy it a third time.

Consider extracting a free function in `types.rs` or a shared module:
```rust
fn extract_parsed_fields(
    frontmatter: Option<DocumentFrontmatter>,
) -> (Option<DocumentOptionsOverride>, Option<DocumentIntegralNameOverride>) {
    let options = frontmatter.as_ref().and_then(|fm| fm.options.clone());
    let legacy_integral = frontmatter
        .and_then(|fm| fm.integral_name_memory)
        .filter(|_| options.as_ref().and_then(|o| o.integral_name_memory.as_ref()).is_none());
    (options, legacy_integral)
}
```

crates/citum-engine/src/processor/document/pipeline.rs
```diff
@@ -36,14 +36,44 @@ impl Processor {
...
+        let owned_org = {
+            let base = owned_integral.as_ref().unwrap_or(self);
+            base.processor_with_document_org_abbreviation_override(effective_org_override)
+        };
+
+        // Apply bibliography overrides from the options block.
+        let owned_bib = parsed
+            .frontmatter_options
+            .as_ref()
+            .filter(|o| o.bibliography.is_some())
```
bdarcus, line 69: **Simplification: `.filter(bibliography.is_some())` is redundant (and the redundancy is the risk)**

`apply_bibliography_to` already early-returns when `self.bibliography` is `None`, so the `.filter()` here is not needed for correctness. It IS useful to avoid the unnecessary `Processor` clone (including bibliography) when there's nothing to apply — but that reasoning lives here while the guard lives there. Two places must stay in sync.

Consider either:
- Dropping the filter and adding the no-op-clone guard inside `processor_with_bibliography_override` (one place to maintain), or
- Adding a comment here: `// guards against cloning a Processor with a large bibliography when no bib override is set`.

crates/citum-cli/src/commands/render/mod.rs
```diff
@@ -53,15 +53,25 @@ pub(super) fn run_render_doc(args: RenderDocArgs) -> CliResult {
         );
     }
 
+    // Read document first so we can extract frontmatter locale for processor setup.
+    let doc_content = fs::read_to_string(&args.input)?;
```
bdarcus, line 57: **Efficiency: frontmatter is parsed twice per `render doc` invocation**

`extract_frontmatter_locale` parses the YAML frontmatter here (serde_yaml allocation + deserialization). Then `process_document` → `parser.parse_document` → `parse_frontmatter` parses it again inside the engine.

This is a known layering constraint (locale must be known before building the `Processor`). Tracking as a future optimization: a `parse_document_locale` in the engine's public API would collapse both parses into one.

crates/citum-cli/src/commands/render/mod.rs
```diff
@@ -307,6 +317,44 @@ fn render_refs_json(
...
+        locale: Option<String>,
+    }
+
+    #[derive(Deserialize)]
+    struct MinimalFrontmatter {
+        options: Option<FrontmatterOptions>,
+    }
+
+    let trimmed = content.trim_start();
+    if !trimmed.starts_with("---") {
```
bdarcus, line 339: **Plausible: `starts_with("---")` has no end-of-line anchor**

`starts_with("---")` matches `---` followed by anything, including `---title: foo` or `---foo`. In those cases `after_opening` is not empty YAML but document body content, and `find("\n---")` may match a later horizontal-rule or code-block delimiter and misparse body text as frontmatter.

The engine's `parse_frontmatter` has the same pattern (it's the model for this function), so this is pre-existing. A defensive fix: `starts_with("---\n") || starts_with("---\r\n")`.

crates/citum-engine/src/processor/document/pipeline.rs
bdarcus: No need to preserve legacy here.

crates/citum-engine/src/processor/document/types.rs
bdarcus: Let's fix this here.

PR #847, opened 2026-05-30
feat(engine): add markdown output format

bdarcus:
## Summary

Closes #844. Supersedes the reverted grid-table preprocessor approach from PR #846.

This PR adds Markdown output as a first-class render target and keeps the supported interop direction explicit: `citum -> pandoc`. Citum processes citation markers and emits Markdown with citation/bibliography text rendered in CommonMark-compatible inline markup, while body-level Markdown is passed through for Pandoc or another downstream Markdown consumer.

It also fixes the Chicago author-date locator regression with the preferred declarative style shape: author and year live in an explicit grouped template with a space delimiter, while the locator owns its comma prefix. The engine now preserves that group delimiter when suppress-author rendering strips the author from the group, instead of relying on a flat year-prefix workaround or a generic spacing heuristic.

## Changes

- Add `--format markdown` support through CLI document/reference rendering.
- Add `render::markdown::Markdown` with CommonMark inline rendering for emphasis, strong, small caps, superscript, quotes, links, and semantic passthrough.
- Escape Markdown-active raw data text, including `*`, `_`, `[`, `]`, backticks, `<`, `>`, and `&`, so bibliography fields do not accidentally become markup or entities downstream.
- Preserve Markdown body blocks verbatim for passthrough output, including pipe tables, fenced code blocks, headings, and footnote syntax.
- Parse Markdown input footnote definitions with `ENABLE_FOOTNOTES`, so citations inside user-authored footnotes render in place as `ManualFootnote` rather than generating duplicate auto-footnotes.
- Enable HTML finalization support for Markdown pipe tables and footnotes.
- Fix grouped author-date citation rendering when author suppression removes the leading grouped child but the remaining year still needs the group delimiter boundary.
- Update Chicago author-date 18th citation layout to use an explicit author/year group and separate locator punctuation.
- Update the Pandoc workflow docs, including the note-style `--input-format markdown` example.

## Pandoc workflow

```bash
citum render doc input.md --input-format markdown --format markdown \
  -s apa -b refs.yaml | pandoc --to html
```

Do not pre-process citation-bearing Markdown with `pandoc --to commonmark`; Pandoc escapes citation marker brackets during that conversion, so Citum can no longer recognize `[@key]` markers.

## Test plan

- [x] Markdown renderer unit tests for inline formatting and text escaping, including `*`, `_`, `[`, `]`, backticks, `<`, `>`, and `&`.
- [x] Markdown parser unit tests for manual footnote citation placement, note order tracking, and multiple footnotes.
- [x] HTML finalization tests for Markdown pipe tables and footnotes.
- [x] Integration coverage for Markdown passthrough with pipe tables, fenced code blocks, citation replacement, and bibliography generation.
- [x] Integration coverage for Chicago notes Markdown footnotes with no duplicate auto-footnotes.
- [x] Chicago author-date parenthetical and suppress-author locator regression tests.
- [x] Engine regression for explicit author+year grouped template with `delimiter: " "`, proving `Author Year` rather than `AuthorYear`.
- [x] Local full gate: `cargo fmt --check && cargo clippy --all-targets --all-features -- -D warnings && cargo nextest run` passed, 1455/1455 tests.
- [x] Pre-push hook gate passed before force-push.

### DIFF
--- a/crates/citum-engine/tests/document.rs
+++ b/crates/citum-engine/tests/document.rs
@@ -746,6 +746,44 @@ fn given_markdown_integral_note_citation_when_rendered_with_a_note_style_then_a_
     );
 }
 
+fn given_markdown_citation_inside_manual_footnote_when_rendered_with_note_style_then_it_renders_in_place()
+ {
+    // When the user writes their own [^n]: block containing a citation,
+    // the citation should render inline inside that definition (ManualFootnote
+    // placement) rather than generating a second auto-footnote.
+    let processor =
+        example_document_processor("styles/embedded/chicago-shortened-notes-bibliography.yaml");
+    let parser = MarkdownParser;
+    let document = "See note[^1].\n\n[^1]: Early work [@kuhn1962] supports this.";
+
+    let output = processor.process_document::<_, citum_engine::render::plain::PlainText>(
+        document,
+        &parser,
+        DocumentFormat::Plain,
+    );
+
+    // The manual footnote anchor must appear in prose.
+    assert!(
+        output.contains("[^1]"),
+        "manual footnote reference should appear in prose: {output}"
+    );
+    // The rendered citation should appear inside the footnote definition, not
+    // as a separate auto-generated note.
+    assert!(
+        output.contains("[^1]: Early work"),
+        "footnote definition body should be preserved: {output}"
+    );
+    assert!(
+        output.contains("Kuhn") || output.contains("Structure"),
+        "citation inside manual footnote should be rendered in place: {output}"
+    );
+    // No auto-generated note should be created for this citation.
+    assert!(
+        !output.contains("[^citum-auto-"),
+        "no auto-footnote should be generated for a ManualFootnote citation: {output}"
+    );
+}
+
 // --- Grouped Bibliography Scenarios ---
 
 fn given_grouped_primary_and_secondary_sources_when_rendered_then_both_group_headings_and_entries_appear()
--- a/crates/citum-engine/tests/document.rs
+++ b/crates/citum-engine/tests/document.rs
@@ -1207,6 +1207,85 @@ mod markdown_documents {
         );
         super::given_markdown_integral_note_citation_when_rendered_with_a_note_style_then_a_generated_note_is_emitted();
     }
+
+    #[test]
+    fn markdown_output_passes_pipe_table_and_code_block_through_verbatim() {
+        announce_behavior(
+            "Markdown output format must pass block-level markup (pipe tables, fenced code blocks) through verbatim while replacing citation markers with rendered inline text.",
+        );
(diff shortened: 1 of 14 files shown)


.beans/archive/csl26-cfrg--fix-chicago-author-date-locator-space.md
bdarcus, line 1: duplicate bean to csl26-6bw9; delete

.beans/archive/csl26-6bw9--fix-chicago-author-date-locator-space.md
```diff
@@ -0,0 +1,10 @@
+---
+# csl26-6bw9
+title: fix Chicago author-date locator space
+status: completed
+type: bug
+created_at: 2026-05-31T01:51:30Z
+updated_at: 2026-05-31T01:51:30Z
+---
+
+Spurious space before locator comma in Chicago author-date Markdown citations (#844). Root cause: whitespace-only delimiter not suppressed when adjacent rendered element begins with a prefix-carried comma. Fixed in render/citation.rs join loop.
```
bdarcus: Again, I don't think this was the right diagnosis; as I pointed out, this was a style bug. See bean csl26-xy94, and align with that (this maybe should be deleted entirely from the PR?

.beans/archive/csl26-xy94--fix-844-correctly-style-level-group-fix-for-chicag.md
bdarcus, line 1: this is an internal details to this PR; remove this bean

crates/citum-engine/tests/document.rs
```diff
@@ -1055,6 +1093,102 @@ mod note_flow {
...
+        output.contains("```rust\nfn hello() {}\n```"),
+        "fenced code block missing or modified: {output}"
+    );
+
+    // Citation marker replaced with rendered inline text.
+    assert!(
+        !output.contains("[@kuhn1962]"),
+        "raw citation marker should be replaced: {output}"
+    );
+    assert!(
```
bdarcus, line 1138: weak assertion

crates/citum-schema-style/embedded/styles/chicago-author-date-18th.yaml
```diff
@@ -104,11 +104,12 @@ citation:
       and-others: et-al
   - date: issued
     form: year
+    prefix: " "
   - variable: locator
     prefix: ", "
   wrap:
     punctuation: parentheses
-  delimiter: " "
+  delimiter: none
```
bdarcus, line 113: I don't understand this. Group should have worked here, and this looks to me like you ignored a bug, and used sub-standard approaches to style design (affixes rather than delimiter).

PR #884, opened 2026-06-08
feat(engine): consolidate bib block rendering

bdarcus:
## Summary

- Adds `render_document_bibliography_blocks` to `grouping.rs` as the single ordered-block loop with one shared `assigned` dedup set — both the fenced-div document pipeline and the `format_document` API become thin wrappers over it, eliminating the two duplicated `HashSet` loops
- Extends `RenderedBibliographyGroup` with `entries: Vec<ProcEntry>` so the API path exposes per-entry metadata without a second render pass
- Adds `BibliographyBlockRequest` / `FormattedBibliographyBlock` types (with JsonSchema derives) and wires `bibliography_blocks` into `FormatDocumentRequest` / `FormatDocumentResult`
- Adds `process_document_with_caller_blocks` to the pipeline for callers that supply `BibliographyGroup` slices directly (e.g. CLI flag) instead of in-document `:::bibliography{...}` fenced divs
- Adds `--bibliography-blocks <json>` to `citum render doc`, using the same `BibliographyBlockRequest` type as the server — closes the gap where the server was wired but the CLI was not
- Fixes `FormatDocumentParams.bibliography_blocks` from `serde_json::Value` to `Vec<BibliographyBlockRequest>` so the JSON Schema entry carries the full type rather than `null`
- Regenerates `docs/schemas/server.json`; updates `SERVER_INTERACTIVE_API.md`

Drops the flawed `feat(server): add bibliography blocks` commit and rebuilds the feature correctly via a single unified primitive.

Closes csl26-o8ji.

## Test plan

- [x] 1525 tests pass (`cargo nextest run`)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean (no `allow` added to new code)
- [x] `docs/schemas/server.json` diff shows `bibliography_blocks` as typed array with `$ref: BibliographyBlockRequest`
- [x] `citum render doc <doc> --bibliography-blocks blocks.json` renders sectional trailing bibliographies
- [x] `citum render doc <doc-with-fenced-divs>` still works unchanged (pipeline path unaffected)


crates/citum-server/src/rpc.rs
```diff
@@ -166,6 +167,8 @@ pub struct FormatDocumentParams {
     pub refs: serde_json::Value,
     /// Ordered citations as they appear in the document.
     pub citations: serde_json::Value,
+    /// Optional bibliography blocks to render in document order.
+    pub bibliography_blocks: Option<Vec<BibliographyBlockRequest>>,
```
bdarcus: **`bibliography_blocks` type mismatch between schema mirror and engine request.**

`FormatDocumentParams` (the server's schema mirror) declares `bibliography_blocks: Option<Vec<BibliographyBlockRequest>>`, but the authoritative `FormatDocumentRequest` uses `Vec<BibliographyBlockRequest>` with `#[serde(default)]`. Clients following the server schema could send `null`, which the engine won't accept. The mirror should match: `#[serde(default, skip_serializing_if = "Vec::is_empty")]` on a `Vec` field. Fixed in this review pass.

crates/citum-engine/src/api/document.rs
```diff
@@ -705,6 +892,86 @@ mod tests {
...
+        };
+
+        let result =
+            format_document_with_style(make_test_style_with_bibliography(), request).unwrap();
+
+        assert_eq!(result.bibliography.entries.len(), 2);
+        assert_eq!(result.bibliography_blocks.len(), 2);
+        assert_eq!(result.bibliography_blocks[0].id, "primary");
+        assert_eq!(result.bibliography_blocks[0].entries[0].id, "primary1891");
+        assert!(
```
bdarcus: **Weak `contains()` assertions with substrings under 30 chars (violates CODING_STANDARDS.md).**

`contains("Archive Notes")` and `contains("Secondary Study")` use 13- and 15-char substrings. The coding standard prohibits `contains()` with substrings under 30 chars because they pass trivially and miss formatting regressions. Since the test style is controlled (title-only template), exact equality is available: `assert_eq!(result.bibliography_blocks[0].content, "Archive Notes")`. Fixed in this review pass — both tests now use exact `assert_eq!`.

PR #913, opened 2026-06-12
test(engine): harden disambiguation soundness

bdarcus:
## Summary
- Adds a DISAMBIGUATION §4 integration test for transliterated-name collision keys in `citations.rs`.
- Moves a subsequent et-al behavior wrapper out of the disambiguation behavior group.
- Updates the soundness ledger to `24/0/0/0` for DISAMBIGUATION and wires repo setup to refresh Codex skill symlinks.

## Validation
- `./scripts/validate-frontmatter.sh --copilot-strict`
- `cargo fmt --check && cargo clippy --all-targets --all-features -- -D warnings && cargo nextest run`
- pre-push hook reran the full Rust gate: 1,567 tests passed

### DIFF
--- a/crates/citum-engine/tests/citations.rs
+++ b/crates/citum-engine/tests/citations.rs
@@ -32,8 +32,9 @@ use citum_schema::{
     grouping::{GroupSort, GroupSortEntry, GroupSortKey, SortKey as GroupSortKeyType},
     options::{
         AndOptions, Config, ContributorConfig, DelimiterPrecedesLast, DisplayAsSort, GivennameRule,
-        IntegralNameContexts, IntegralNameMemoryConfig, IntegralNameScope, NameForm, Processing,
-        ProcessingCustom, ShortenListOptions, SubsequentNameForm,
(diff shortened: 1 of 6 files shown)


crates/citum-engine/tests/citations.rs
```diff
@@ -1190,6 +1183,70 @@ fn disambiguation_multilingual_contributors_collide_on_original_family_name() {
...
+                    id: "ml-tanaka-b".to_string(),
+                    ..Default::default()
+                },
+            ],
+            mode: CitationMode::NonIntegral,
+            ..Default::default()
+        })
+        .expect("citation should render");
+
+    assert_eq!(result, "Tanaka, (2020a); Tanaka, (2020b)");
```
bdarcus: I may not be tracing exactly what's going on here, but at face value, isn't this wrong? They are different authors, no? So then there should be no year suffix for either, and the given name (or, depending on the styles config, initials) should be added?

If that's correct, rewrite the test with the correct assertion, and fix the failing test on a new commit that precedes this commit.

PR #920, opened 2026-06-13
Disambiguation migration presets

bdarcus:
## Summary

- lock the csl26-1861 disambiguation migration decisions in the bean and reference docs
- add public author-date processing variants for B1/givenname/names/full disambiguation presets
- make bare `author-date` use the B1 default and fold exact explicit CSL shapes to named presets
- add an engine regression test for year-suffix-only author-date disambiguation

## Tests

- `cargo test -p citum-schema-style processing_author_date`
- `cargo test -p citum-migrate extract_author_date_processing`
- `cargo test -p citum-migrate processing_disambiguation`
- `cargo test -p citum-migrate extract_givenname_rule`
- `cargo test -p citum-engine test_author_date_default_uses_year_suffix_without_name_expansion`
- `cargo run --bin citum --features schema -- schema --out-dir docs/schemas`
- `cargo fmt --check && cargo clippy --all-targets --all-features -- -D warnings && cargo nextest run`
- `node scripts/report-migrate-sqi.js --corpus random --sample 100 --seed 20260610`
- `node scripts/report-core.js > /tmp/r.json && node scripts/check-core-quality.js --report /tmp/r.json --baseline scripts/report-data/core-quality-baseline.json`

Core quality passed with two existing reporter warnings recorded in `/tmp/r.json`.

### DIFF
--- a/crates/citum-migrate/src/citation_validate.rs
+++ b/crates/citum-migrate/src/citation_validate.rs
@@ -109,10 +109,10 @@ fn normalize_note_or_author_date_citation(
     style_name: &str,
 ) {
     let should_normalize = legacy_style.class == "note"
-        || matches!(
-            options.processing,
-            Some(citum_schema::options::Processing::AuthorDate)
-        );
+        || options
+            .processing
+            .as_ref()
+            .is_some_and(citum_schema::options::Processing::is_author_date_family);
     if !should_normalize {
         return;
     }
--- a/crates/citum-migrate/src/options_extractor/processing.rs
+++ b/crates/citum-migrate/src/options_extractor/processing.rs
@@ -4,8 +4,7 @@ SPDX-FileCopyrightText: © 2023-2026 Bruce D'Arcus and Citum contributors
 */
 
 use citum_schema::options::{
-    Disambiguation, GivennameRule, Group, Processing, ProcessingCustom, Sort, SortEntry, SortKey,
-    SortSpec,
+    GivennameRule, Group, Processing, ProcessingCustom, Sort, SortEntry, SortKey, SortSpec,
 };
 use citum_schema::presets::SortPreset;
 use csl_legacy::model::{CslNode, Style};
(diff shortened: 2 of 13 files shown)


crates/citum-migrate/src/options_extractor/processing.rs
```diff
@@ -52,38 +51,33 @@ pub fn detect_processing_mode(style: &Style) -> Option<Processing> {
...
-        };
+        let mut custom = Processing::AuthorDate.config();
 
         let sort = style.citation.sort.as_ref().and_then(extract_sort);
-        let group = sort
-            .as_ref()
-            .map(SortEntry::resolve)
-            .and_then(|sort| extract_group_from_sort(&sort));
+        if let Some(sort) = sort {
+            custom.group = extract_group_from_sort(&sort.resolve());
```
bdarcus: **Bug: preset sorts overwrite the base-config group with a 3-key resolved sort, breaking fold back to named variants.**

`sort.resolve()` on `SortEntry::Preset(AuthorDateTitle)` returns `[Author, Year, Title]` (3 keys), but `author_date_config` sets `Group { template: [Author, Year] }` (2 keys). The mismatch means `fold_to_named_processing` can never match any `AuthorDate*` variant for styles with an explicit sort element — even one that maps cleanly to the preset — so they silently emit `Processing::Custom` instead of the idiomatic named form.

Fixed in this review: skip `extract_group_from_sort` when the sort is a `SortEntry::Preset`; the base config already carries the canonical group.

crates/citum-schema-style/src/options/processing.rs
```diff
@@ -214,18 +259,10 @@ impl Processing {
...
-                    names: true,
-                    add_givenname: true,
-                    givenname_rule: GivennameRule::default(),
-                    year_suffix: true,
-                }),
-            },
+            Processing::AuthorDate => author_date_config(false, false),
+            Processing::AuthorDateGivenname => author_date_config(false, true),
+            Processing::AuthorDateNames => author_date_config(true, false),
+            Processing::AuthorDateFull => author_date_config(true, true),
```
bdarcus, line 268: **Breaking change to engine fallback for `processing: None` styles (not fixed here — design decision needed).**

`Processing::AuthorDate.config()` now returns `names: false, add_givenname: false` instead of the old `true/true`. Three engine sites use this as a fallback when `processing` is absent:

- `crates/citum-engine/src/processor/disambiguation.rs` — `None` arm uses `Processing::AuthorDate.config().disambiguate`
- `crates/citum-engine/src/processor/citation.rs` — same pattern

126 YAML styles in `styles/` omit the `processing:` field. They previously inherited `names: true, add_givenname: true` from the old default; they now silently get `names: false, add_givenname: false`, suppressing all name-based disambiguation without any error or migration prompt.

The `year_suffix` field (still `true`) is unaffected. **Decide:** should `processing: None` mean «conservative year-suffix-only» (keep current engine fallback) or «legacy-compatible» (change engine fallback to `AuthorDateFull`)? The engine fallback sites need an explicit update.

crates/citum-migrate/src/citation_validate.rs
```diff
@@ -111,7 +111,12 @@ fn normalize_note_or_author_date_citation(
     let should_normalize = legacy_style.class == "note"
         || matches!(
             options.processing,
-            Some(citum_schema::options::Processing::AuthorDate)
+            Some(
+                citum_schema::options::Processing::AuthorDate
```
bdarcus: **Simplification: three non-test sites spell out all 4 author-date variants with no compiler enforcement.**

`citation_validate.rs`, `compilation.rs`, and `processing.rs:default_bibliography_sort` each have their own 4-arm OR-pattern. There is no `Processing::is_author_date_family()` predicate, so adding a 5th variant requires updating every site by hand — with no compiler error if one is missed.

Fixed in this review: added `pub fn is_author_date_family(&self) -> bool` to `Processing` and replaced the `matches!` blocks in `citation_validate.rs` and `compilation.rs` with `.is_some_and(Processing::is_author_date_family)`.

crates/citum-schema-style/src/options/processing.rs
```diff
@@ -20,6 +20,27 @@ use serde::{Deserialize, Serialize};
...
+];
+
+const PROCESSING_ENUM_VARIANTS: &[&str] = &[
+    "author-date",
+    "author-date-givenname",
+    "author-date-names",
+    "author-date-full",
+    "numeric",
+    "note",
+    "label",
```
bdarcus: **Maintenance risk: two parallel `&[&str]` variant arrays differ by exactly one entry (`"custom"`).**

`PROCESSING_STRING_VARIANTS` (7 entries) and `PROCESSING_ENUM_VARIANTS` (8 entries) must be kept in sync manually — adding a new named variant requires updating both. There is no compiler enforcement. The two arrays are each consumed exactly once, only in `unknown_variant` error paths.

Cheapest fix: express `PROCESSING_ENUM_VARIANTS` as a comment note that it is `PROCESSING_STRING_VARIANTS + "custom"`, or switch to a `strum::VariantNames` derive to generate them automatically. Not fixed here, but worth tracking.

crates/citum-schema-style/src/options/processing.rs
bdarcus: There should not be idiosyncratic settings like this; the default should be set once, and then propagate. E.g. if IUC, this sounds like a bug.

crates/citum-schema-style/src/options/processing.rs
bdarcus: Fix this now.

crates/citum-schema-style/src/options/processing.rs
bdarcus: Settled — and it's not a regression. CSL/citeproc-js default `disambiguate-add-names`/`-add-givenname` to **false**, so `false/false` is the faithful default; the old `true/true` was latent over-disambiguation. Fixed by centralizing: added `Config::effective_processing()` → `Processing::default()` (AuthorDate: year-suffix on, name-escalation off), and routed all three engine sites (disambiguation.rs, citation.rs, date.rs) through it. One source, propagates. Scorecard confirms: **fidelity=1.0 across 154 styles, 0 warnings**.

crates/citum-schema-style/src/options/processing.rs
bdarcus: Fixed. Deleted `PROCESSING_ENUM_VARIANTS`; `visit_enum` now uses `&["custom"]` (the only externally-tagged variant — the string forms are handled by `visit_str`). `PROCESSING_STRING_VARIANTS` is the single remaining source.

crates/citum-migrate/src/options_extractor/processing.rs
bdarcus: Fixed in the folded `feat` commit: `detect_processing_mode` now skips `extract_group_from_sort` for `SortEntry::Preset` (the base config already carries the canonical `[Author, Year]` group), so preset-sort styles fold to the named variant instead of `Custom`. Covered by the updated duplicate-macro-keys test.

crates/citum-migrate/src/citation_validate.rs
bdarcus: Fixed. Added `Processing::is_author_date_family()`; `citation_validate.rs` and `compilation.rs` now call it instead of spelling out the 4-arm `matches!`.

PR #932, opened 2026-06-16
fix(migrate): improve authorable compression

bdarcus:
## Summary
- Fold exact substitute configs to named presets during migration output
- Strip inert suppressed placeholders before bibliography variant serialization
- Bias type-variant compression toward maintainer-readable parents and grouped selectors
- Add an ACME regression covering the authorability shape

## Rationale
The migrated YAML for styles like ACME was structurally valid but hard for style maintainers to read: standard substitute chains were expanded, suppressed term-only groups survived as dead anchors, and rare type variants became surprising parents for common types. This keeps fidelity as the hard gate while making emitted Citum styles more authorable.

## Validation
- `cargo test -p citum-migrate`
- `cargo fmt --check`
- `just pre-commit`
- pre-push hook: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo nextest run`

### DIFF
--- a/crates/citum-migrate/src/template_diff.rs
+++ b/crates/citum-migrate/src/template_diff.rs
@@ -13,19 +13,20 @@ use citum_schema::{
         TypeSelector,
     },
 };
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 
 /// Per-type template map used throughout the migration pipeline.
 pub(crate) type TypeTemplateMap = indexmap::IndexMap<TypeSelector, Vec<TemplateComponent>>;
 
 pub(crate) type TypeVariantMap = indexmap::IndexMap<TypeSelector, TemplateVariant>;
 
+const COMMON_PARENT_MIN_EXTENDS_SAVINGS: usize = 1;
+const RARE_PARENT_MIN_EXTENDS_SAVINGS: usize = 2;
+
 /// Builds all template variants for a given reference type from a type-template map.
 ///
-/// This runs before lineage/wrapper conversion, so raw template equality is not
-/// final semantic equality. Preserve the original selector keys here; any
-/// selector grouping must happen only after resolved parent-aware templates are
-/// known.
+/// This preserves addressable parent selectors for `extends` resolution while
+/// grouping unreferenced selectors whose intended full templates are identical.
 pub(crate) fn build_type_variants(
     default_template: &[TemplateComponent],
     type_templates: TypeTemplateMap,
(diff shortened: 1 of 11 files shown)


crates/citum-migrate/src/template_diff.rs
```diff
@@ -158,6 +168,122 @@ fn diff_operation_weight(diff: &TemplateVariantDiff) -> usize {
...
+fn extends_authoring_penalty(parent_selector: &TypeSelector, selector: &TypeSelector) -> usize {
+    let selector_has_type = |candidate: &TypeSelector, type_name: &str| {
+        selector_type_names(candidate)
+            .iter()
+            .any(|name| name.replace('_', "-") == type_name)
+    };
+    let parent_rank = selector_authoring_rank(parent_selector);
+    let child_rank = selector_authoring_rank(selector);
+    let rare_parent_penalty = if parent_rank > child_rank { 3 } else { 0 };
+    let webpage_encyclopedia_penalty = if selector_has_type(selector, "webpage")
```
bdarcus, line 180: **Altitude (low):** the `webpage`/`entry-encyclopedia` pair is a hard-coded special case layered on the general authoring-rank mechanism. I verified empirically that it's redundant for the ACME regression: neutralizing this branch (`= 0`) keeps `acme_migration_emits_authorable_compressed_yaml` green, because the general `rare_parent_penalty` already fires here (`parent_rank 3 > child_rank 1` → +3, plus `parent_rank` 3 = 6 of penalty before this `+4`).

Not a bug — it's an extra safety margin — but the next undesired extends pair will need another hard-coded branch. Worth a comment explaining why 6 isn't enough margin, or generalizing (e.g. always penalize when `parent_rank - child_rank >= 2`).

crates/citum-migrate/src/template_diff.rs
```diff
@@ -158,6 +168,122 @@ fn diff_operation_weight(diff: &TemplateVariantDiff) -> usize {
...
+        RARE_PARENT_MIN_EXTENDS_SAVINGS
+    }
+}
+
+fn selector_authoring_rank(selector: &TypeSelector) -> usize {
+    selector_type_names(selector)
+        .iter()
+        .map(|name| type_authoring_rank(name))
+        .min()
+        .unwrap_or(usize::MAX)
```
bdarcus: **Defensive (low):** for an empty `TypeSelector::Multiple(vec![])` this returns `usize::MAX`, which then flows into `parent_rank + rare_parent_penalty + webpage_encyclopedia_penalty` (and `raw_weight + extends_authoring_penalty(...)` in `derive_best_template_variant_diff`) and panics on overflow in debug builds.

Unreachable with current selector construction (selectors always carry ≥1 type name), so it's latent rather than active. A finite fallback — e.g. `.unwrap_or(4)` (the unknown-type rank) — removes the foot-gun with identical behavior for every reachable input.

PR #965, opened 2026-06-24
Add localized message components, deprecate `term`

bdarcus:
## Summary
- add `TemplateMessage` / `message:` style-template component for locale-authored MF2 phrase calls
- extend locale message args/resolution and render message args through normal component rendering
- document the schema/template boundary, seed initial phrase IDs in embedded locales, and regenerate schemas

## Checks
- `cargo test -p citum-schema-style --lib -- --nocapture`
- `cargo test -p citum-engine --lib test_message_component -- --nocapture`
- `just schema-gen`
- `just pre-commit`
- pre-push hook reran fmt, clippy, and nextest before push

### DIFF
--- a/.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
+++ b/.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
@@ -1,7 +1,7 @@
 ---
 # csl26-fdzc
 title: Migrate styles to localized message phrase calls
-status: todo
+status: in-progress
 type: task
 priority: high
 tags:
@@ -9,7 +9,7 @@ tags:
     - localization
     - mf2
 created_at: 2026-06-25T00:22:34Z
-updated_at: 2026-06-25T00:22:34Z
+updated_at: 2026-06-25T10:16:23Z
 ---
 
 ## Problem
@@ -30,8 +30,8 @@ through the rest of `styles/` by style family or shared template pattern.
   `message:` calls where the locale should control word order.
 - Prioritize `pattern.accessed-date`, `pattern.in-container`,
   `pattern.available-at`, and `pattern.retrieved-from`.
-- Keep atomic labels as `term.*` locale messages where they are labels rather
-  than compositional phrases.
+- Keep lexical and inflectional labels as `term.*` or `role.*` locale messages
+  where they are labels rather than phrase realization.
 
 ## Acceptance Criteria
 
@@ -40,3 +40,12 @@ through the rest of `styles/` by style family or shared template pattern.
 - Missing message IDs and missing message args are caught by lint before merge.
 - Embedded styles are completed before broad non-embedded migration begins.
 - Any new phrase IDs are documented in `docs/specs/LOCALE_MESSAGES.md`.
+
+
+## Embedded Proof Batch (PR #965)
+
+- Started the embedded migration with output-equivalent `message:` calls in embedded-core styles.
+- Converted representative `pattern.accessed-date` call sites in AMA, Chicago, Elsevier, MLA, and Springer Vancouver styles.
+- Converted representative `pattern.in-container` call sites in Chicago, IEEE, and Springer author-date styles.
+- Added grouped message arguments so `pattern.in-container` can receive an already-rendered container cluster such as editor plus parent-monograph title.
+- Deferred APA's container-author site, colon-bearing `in:` sites, `URL ` labels, and role-plus-name phrases until later batches define phrase IDs that preserve those outputs without encoding English glue inside arguments.
--- a/.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
+++ b/.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
@@ -9,7 +9,7 @@ tags:
     - localization
     - mf2
 created_at: 2026-06-25T00:22:34Z
-updated_at: 2026-06-25T10:16:23Z
+updated_at: 2026-06-25T11:02:06Z
 ---
 
 ## Problem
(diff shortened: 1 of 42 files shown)


docs/specs/LOCALE_MESSAGES.md
```diff
@@ -123,6 +124,52 @@ parameterized if and only if its body contains a `{` character.
...
+The template-schema `term:` component remains parseable for compatibility but
+is deprecated for new phrase realization. This does not remove `term.*` message
+IDs from locale files; those IDs remain the canonical representation for atomic
+labels.
+
+```yaml
+# Locale file
+messages:
+  pattern.accessed-date: "accessed {$date}"
+  pattern.in-container: "in {$container}"
```
bdarcus, line 169: "in" here is basically a prefix for a group of variables (that represents an edited book, for example), rather than a single. Does it matter? If it does, what do we do about it? 

I think the answer is largely empirical; we should analyze existing CSL and Citum styles.

docs/specs/LOCALE_MESSAGES.md
```diff
@@ -123,6 +124,61 @@ parameterized if and only if its body contains a `{` character.
 
 ---
 
+### 1.4.1 Lexical Labels vs Phrase Realization
+
+Locale messages have two different jobs, but the boundary is behavioral rather
+than a hard namespace split:
+
+- **Lexical or inflectional labels** (`term.*`, `role.*`) name a localizable
```
bdarcus, line 138: Define these terms, for those new to localization.

docs/specs/LOCALE_MESSAGES.md
```diff
@@ -123,6 +124,61 @@ parameterized if and only if its body contains a `{` character.
...
+arguments; the active locale decides the phrase order and glue text.
+
+Role-related messages can belong to either side of this distinction. A suffix
+label like `role.editor.label` is lexical or inflectional. A role-plus-name
+phrase such as `edited by {$names}` is phrase realization and may be modeled as
+a `pattern.*` message when a style needs the locale to control placement or word
+order around the rendered names.
+
+The template-schema `term:` component remains parseable for compatibility but
+is deprecated for new phrase realization. This does not remove `term.*` message
```
bdarcus: strengthen: `term:` will be removed as soon as the existing styles are upgraded to remove it.

docs/specs/LOCALE_MESSAGES.md
```diff
@@ -123,6 +124,61 @@ parameterized if and only if its body contains a `{` character.
...
+  `pattern.in-container`, `pattern.available-at`, `pattern.retrieved-from`.
+
+Styles SHOULD use `message:` template components for compositional phrases
+instead of spelling English glue with `term:` components, `prefix`, or `suffix`.
+The style still decides *which* phrase is needed and *which fields* supply its
+arguments; the active locale decides the phrase order and glue text.
+
+Role-related messages can belong to either side of this distinction. A suffix
+label like `role.editor.label` is lexical or inflectional. A role-plus-name
+phrase such as `edited by {$names}` is phrase realization and may be modeled as
```
bdarcus: given this (sensible) guidance, don't we need to add, and use, a message ID for this?

.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
```diff
@@ -0,0 +1,94 @@
...
+- Added grouped message arguments so `pattern.in-container` can receive an
+  already-rendered container cluster such as editor plus parent-monograph title.
+- Directly touched 9 embedded files. Wrapper styles may inherit those changes,
+  but remaining embedded-core files still need classification and migration.
+- Remaining embedded candidates include `elsevier-with-titles-core`,
+  `springer-basic-brackets-core`,
+  `taylor-and-francis-council-of-science-editors-author-date-core`,
+  `taylor-and-francis-national-library-of-medicine-core`, and residual
+  `term: in` sites in already-touched families.
+- Deferred APA's container-author site, Chicago author-date's German override
```
bdarcus: is this really deferred at this point?

docs/specs/LOCALE_MESSAGES.md
bdarcus: Treating as already-resolved per your note — no change. (For the record: the embedded styles already pass `$container` as either a single field or a rendered group/cluster, e.g. IEEE = parent-monograph alone vs APA = container-author + parent-monograph, so the grouped-arg path already covers the edited-book case.)

docs/specs/LOCALE_MESSAGES.md
bdarcus: Done in f8f8d0d2 — added a glossary sentence before the bullets defining *lexical* (the localized word/abbrev a language picks), *inflectional* (that word varied by count/gender), and *phrase realization* (assembling rendered values with the locale's glue + word order).

docs/specs/LOCALE_MESSAGES.md
bdarcus: Strengthened in f8f8d0d2 — now reads that `term:` is deprecated and "will be removed as soon as the existing styles are upgraded to drop it," while keeping the note that `term.*` message IDs stay canonical for labels/inflected roles.

docs/specs/LOCALE_MESSAGES.md
bdarcus: Addressed in f8f8d0d2 — the example now names a concrete planned ID, `pattern.editor-contribution`, and states that introducing/adopting it is deferred to the role-plus-name batch tracked in bean csl26-fdzc (no style calls it yet). That keeps the spec consistent with the bean's deferral rather than shipping the ID now.

.beans/csl26-fdzc--migrate-styles-to-localized-message-phrase-calls.md
bdarcus: Good catch — reconciled in a499fd49. APA's container-author site and Chicago author-date's German-override container site were completed in the "APA And Chicago Follow-up" section, so they're dropped from the Deferred list (now: colon-bearing `in:` sites, `URL ` labels, role-plus-name phrases) with a cross-reference noting the two were finished.

PR #1008, opened 2026-07-05
fix(engine): centralize type classification (spec + Part A)

bdarcus:
## Summary
- Revised the spec per review: reframed away from citeproc-matching as a goal (prior art only), clarified the new `TypeLabel` component emits term text only (brackets come from style-level `wrap: brackets`, not the component), and resolved all four open decisions into settled Design Decisions. Status flipped Draft → Active.
- Implemented **Part A**: new `crates/citum-engine/src/values/type_class.rs` module centralizes 5 of the 6 audit-flagged hardcoded per-`ref_type` sites (title category for Primary/ContainerTitle/ParentSerial, `TypeClass` legal/classical membership, serial-parent gating, `chapter`→`entry-dictionary` type-variant alias, dataset DOI-URL synthesis). Each site is now a thin call-through instead of a scattered `match`/`contains` arm.
- Confirmed the `contains("ancient")` fuzzy match (Finding 14) matches nothing the data model can produce — `ref_type()` is a finite, enumerable string set and the `Classic` class always emits the literal `"classic"` — and replaced it with an explicit member list.
- **Part B** (new `TypeLabel` component replacing the `[Dataset]` suffix hack, `feat`, schema-version bump) is tracked separately as `csl26-rvhe`, blocked by this PR.

## Test plan
- [x] `cargo nextest run -p citum-engine`: 898/898 pass
- [x] `node scripts/oracle.js styles-legacy/apa.csl`: identical 45/46 result vs. `main` (same pre-existing failure signature) — zero behavior drift
- [x] Full workspace `just pre-commit`: 1774/1774 tests, fmt, clippy `-D warnings` clean
- [x] `./scripts/validate-frontmatter.sh --copilot-strict` passes on the revised spec
- [x] No `contains("article")`/`contains("ancient")` per-`ref_type` matching remains at the five sites (grep-verified)


docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
```diff
@@ -0,0 +1,183 @@
...
+## Open Decisions
+
+1. **Sequencing.** (a) Two PRs — land Part A (schema-neutral) first, then
+   Part B as a `feat`; or (b) one combined PR. Recommendation: **(a)**, so
+   the mechanical centralization is not gated on the type-label design.
+2. **Type-label component shape for Part B.** Options:
+   - A new dedicated `TemplateComponent` (e.g. `TypeLabel`) that resolves
+     the reference-type term with genre/medium fallback and standard
+     wrap/text-case rendering. Cleanest semantic match to citeproc's
+     `description` macro; largest schema surface.
```
bdarcus: Matching citeproc is not our primary goal; it is a sound internal design and implementation. Right?

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
```diff
@@ -0,0 +1,183 @@
...
+   - Extend `TemplateComponent::Term` with a "term = this reference's type"
+     selector plus a fallback list, reusing existing term rendering. Smaller
+     surface; overloads `Term`.
+   - Keep the current title-suffix mechanism but source the label from a
+     locale term instead of the literal string (interim, removes
+     locale-coupling only; still a de-dup hack). Smallest; least correct.
+3. **`type_class` module home.** `values/type_class.rs` (next to today's
+   `type_class_matches`) vs. a crate-root `type_class.rs`. Recommendation:
+   **`values/type_class.rs`**.
```
bdarcus: that's fine

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
```diff
@@ -0,0 +1,183 @@
...
+own type* (with the genre/medium fallback) rather than a fixed term name.
+
+The dataset type-variant in `apa-7th.yaml` becomes, conceptually:
+
+```yaml
+dataset:
+- contributor: author   # unchanged
+- date: issued          # unchanged
+- title: primary        # no more " [Dataset]." suffix
+- <type-label>          # new: localized term, wrapped in brackets
```
bdarcus:
```suggestion
- <type-label>          # new: localized term
```

If we add this to the template, we wouldn't be hard-coding brackets; no?

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
```diff
@@ -0,0 +1,183 @@
...
+     surface; overloads `Term`.
+   - Keep the current title-suffix mechanism but source the label from a
+     locale term instead of the literal string (interim, removes
+     locale-coupling only; still a de-dup hack). Smallest; least correct.
+3. **`type_class` module home.** `values/type_class.rs` (next to today's
+   `type_class_matches`) vs. a crate-root `type_class.rs`. Recommendation:
+   **`values/type_class.rs`**.
+4. **`contains("ancient")` fate.** Replace with an explicit member list
+   (e.g. `ancient`, `ancient-text`?) — needs the actual set of classical
+   ref_types Citum recognizes. Requires confirming the type vocabulary
```
bdarcus: Correct

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
bdarcus: Reframed. Purpose and Design Part B now state citeproc's `[Dataset]` behavior is prior art, not a target — Citum's fix is justified on its own (localized, term-driven, no English literals) independent of matching citeproc output. See v1.1 Purpose + Design Part B intro.

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
bdarcus: Agreed — updated. `TypeLabel` emits only the resolved term text; brackets come from the existing `wrap: brackets` rendering option (`Rendering.wrap`), same as any other component. Nothing is hardcoded in the component. Spec + YAML example updated accordingly, and this shape is now implemented as `TemplateComponent::TypeLabel` with standard `Rendering` in the follow-up Part B work (csl26-rvhe).

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
bdarcus: Confirmed — `values/type_class.rs` is where Part A landed in this PR.

docs/specs/TYPE_CLASSIFICATION_CENTRALIZATION.md
bdarcus: Confirmed and verified: `ref_type()` (citum-schema-data accessors.rs) emits a finite CSL-string set, and the `Classic` class always emits the literal "classic" — never a string containing "ancient". Replaced with an explicit member list (`classic`, `religious-text`) in `matches_type_class`; a test documents the dead fuzzy match had no live producer.

PR #1011, opened 2026-07-05
Configure structured title delimiters

bdarcus:
## Summary

Adds style-level structured title delimiter configuration for GitHub issue #1010.

- Adds `primary-delimiter` for the main-title-to-first-subtitle boundary.
- Adds `subtitle-delimiter` for boundaries between multiple subtitle parts.
- Preserves the existing `": "` default behavior when neither option is set.
- Allows existing title `locale-overrides` to override either delimiter.
- Documents the behavior in `docs/specs/TITLE_TEXT_CASE.md` and regenerates `docs/schemas/style.json`.
- Tracks terminal-punctuation delimiter suppression as follow-up bean `csl26-zfqr`.

## Scope

Touches title rendering configuration and structured title rendering only. Does not add per-template title delimiter overrides and does not introduce an MF2 locale-message path for title assembly.

## Validation

- `just schema-gen`
- `cargo test -p citum-schema-style test_title_locale_overrides_deserialization`
- `cargo test -p citum-engine structured_title`
- `just pre-commit` (1779 tests passed)
- pre-push gate reran fmt, clippy, and nextest (1779 tests passed)

## Risk

Low. Defaults preserve existing structured title output, and new behavior is opt-in via style title config. The main open edge case is significant terminal punctuation before a subtitle delimiter, tracked separately in `csl26-zfqr`.

Closes #1010

### DIFF
--- a/crates/citum-engine/src/values/title.rs
+++ b/crates/citum-engine/src/values/title.rs
@@ -15,7 +15,8 @@ use crate::values::text_case::{self, apply_text_case, capitalize_first_word};
 use crate::values::{
     ComponentValues, ProcHints, ProcValues, RenderOptions, effective_component_language,
 };
-use citum_schema::options::titles::TextCase;
+use citum_schema::locale::Locale;
+use citum_schema::options::titles::{TextCase, TitleRendering};
 use citum_schema::reference::ClassExtension;
 use citum_schema::reference::types::{StructuredTitle, Subtitle, Title};
 use citum_schema::template::{TemplateComponent, TemplateTitle, TitleForm, TitleType};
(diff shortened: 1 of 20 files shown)


docs/specs/TITLE_TEXT_CASE.md
```diff
@@ -119,6 +121,41 @@ recognize at least these punctuation boundaries:
...
+- Delimiter selection is style-owned. Locale-specific delimiter differences
+  may be expressed through existing title `locale-overrides`; this spec does
+  not require an MF2 message path for structured title assembly.
+
+For example:
+
+```yaml
+options:
+  titles:
+    monograph:
```
bdarcus, line 157: how does one set these for ALL titles?

crates/citum-engine/src/values/title.rs
```diff
@@ -230,21 +235,26 @@ fn render_structured_title<F: crate::render::format::OutputFormat<Output = Strin
         other => other,
     });
 
-    let mut parts = vec![main_rendered];
+    let mut rendered = main_rendered;
```
bdarcus: question: why aren't we using `join()` in this function a t all?

PR #1050, opened 2026-07-13
docs(spec): cross-role contributor lists design

bdarcus:
## Summary

Design spec (Draft) for **merged multi-role contributor lists** — bean
`csl26-7ip9`, upstream [CSL schema#442](https://github.com/citation-style-language/schema/issues/442).

Citum currently has no way to render contributors from multiple roles as one
combined name list (APA 7 TV episode: `Kogen, J. (Writer), Wolodarsky, W.
(Writer), & Kirkland, M. (Director).`). CSL 1.0's `editortranslator` virtual
variable is a hardcoded single-pairing workaround with known limitations
(non-generalizable, whole-list identity only, position conflicts).

This PR is **spec-only**; implementation follows in later commits on this
branch after spec review.

## Design highlights

- `TemplateContributor.contributor` accepts a single role (unchanged) or an
  ordered role list; a `merge:` block (list form only) declares:
  - `order: document | role` — credit-order interleaving (APA) vs role-grouped
    topology (MLA); orthogonal to label mode
  - `labels: individual | collective | none` with per-role overrides
    (APA's writer/director-individual vs producers-collective asymmetry)
  - `combine-same-person` (default true) — per-person identity detection,
    fixing `editortranslator`'s whole-list limitation
  - `role-conjunction` — verbatim connector override for composed
    combined-label fallback
- Same-person detection: exact field-wise name equality after NFC + trim;
  structured vs literal never match; no fuzzy matching (no person IDs in the
  data model — same contract citeproc-js uses).
- Combined labels resolve authored locale terms first (hyphenated keys;
  first consumer of the already-shipped en-US `editor-translator` terms),
  composed fallback with a locale/component role-term connector kept
  independent of the name-list conjunction.
- Same-person elision across components (APA: omit `[Recorded by …]` when
  songwriter == performer) via new `render-when` fields
  `contributors-distinct` / `contributors-identical` (distinct = not
  identical, not disjoint).
- Et-al shortening counts entries after combination.
- Migration mapping for CSL 1.0 `<names variable="editor translator">` and
  the `editortranslator` term.

Out of scope: author-slot repositioning (existing substitute machinery),
`CitationCollapse`, new role-taxonomy values (`csl26-013w`).

## Evidence

- Triage source: `docs/architecture/audits/2026-07-12_CSL_SCHEMA_ISSUE_TRIAGE.md`
  (bucket 2, #442 — genuine gap).
- Spec incorporates one external review round (4 blocking, 4 non-blocking
  findings resolved; one reviewer claim corrected — partial-overlap person
  sets satisfy `contributors-distinct` by design).
- Docs-only commit: frontmatter preflight passes; no Rust gate applicable.

This is a broader solution for https://github.com/citation-style-language/schema/issues/442

### DIFF
--- a/docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
+++ b/docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
@@ -1,0 +1,349 @@
+# Cross-Role Contributor Lists Specification
+
+**Status:** Draft
+**Version:** 0.1
+**Date:** 2026-07-13
+**Supersedes:** None
+**Related:** bean `csl26-7ip9`,
+[CSL schema#442](https://github.com/citation-style-language/schema/issues/442),
+[`SECONDARY_CONTRIBUTOR_ROLE_FORMATTING.md`](./SECONDARY_CONTRIBUTOR_ROLE_FORMATTING.md),
+[`ROLE_LABEL_DEFAULTS.md`](./ROLE_LABEL_DEFAULTS.md),
+[`ROLE_SUBSTITUTE_FALLBACK.md`](./ROLE_SUBSTITUTE_FALLBACK.md),
+[`CONTRIBUTOR_PHRASE_MESSAGES.md`](./CONTRIBUTOR_PHRASE_MESSAGES.md)
+
+## Purpose
+
+Define how a style renders contributors from **multiple roles as one merged name
+list** — interleaved entries with per-name or per-group role labels, and a single
+combined-label entry when the same person holds more than one role. This is the
+generalization CSL 1.0 never had: its hardcoded `editortranslator` virtual
+variable covers exactly one role pairing, fires only on whole-list identity, and
+cannot express per-name labels. The motivating requirements come from APA 7
+multimedia references, which need all three behaviors at once:
+
+```
+Kogen, J. (Writer), Wolodarsky, W. (Writer), & Kirkland, M. (Director). (1992).
+Whedon, J. (Writer & Director). (2003).
+```
+
+## Scope
+
+In scope:
+
+- Template schema: `contributor:` accepting an ordered role list, plus a `merge:`
+  configuration block (ordering, label modes, per-role overrides, same-person
+  combination).
+- Rendering semantics: entry ordering, individual/collective/none labeling,
+  same-person detection and combined-label resolution.
+- Locale model for combined-role terms and the role-term connector.
+- An options-level same-person role-suppression rule for elision across
+  components.
+- Migration mapping from CSL 1.0 `<names variable="editor translator">` and the
+  `editortranslator` term.
+
+Out of scope:
+
+- Repositioning contributors into the author slot (e.g. MLA's editor-translator
+  as primary entry) — that is substitution, covered by
+  [`ROLE_SUBSTITUTE_FALLBACK.md`](./ROLE_SUBSTITUTE_FALLBACK.md).
+- `CitationCollapse` — the citation-cluster collapse mechanism is unrelated;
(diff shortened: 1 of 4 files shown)


docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
```diff
@@ -0,0 +1,335 @@
...
+```yaml
+- group:
+  - text: recorded-by-descriptor   # illustrative component
+  render-when:
+    contributors-distinct: [composer, performer]
+```
+
+- `contributors-distinct: [r₁, r₂]` — true iff both roles resolve to non-empty
+  person sets and the sets are **not** equal.
+- `contributors-identical: [r₁, r₂]` — true iff both roles resolve to non-empty
```
bdarcus: is not `iff` a type in both cases?

docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
```diff
@@ -0,0 +1,335 @@
...
+label mechanism can express, because the suppressed text lives in a different
+template component. Two new optional fields on `TemplateGroupCondition`
+(`render-when`) cover it:
+
+```yaml
+- group:
+  - text: recorded-by-descriptor   # illustrative component
+  render-when:
+    contributors-distinct: [composer, performer]
+```
```
bdarcus: I kind if missed that we added `render-when` to the template part of the schema, which I should have pushed back on more earlier. Why do we need this in general, and specifically here? Why not handle it in options?

docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
bdarcus: Intentional — `iff` as "if and only if" — but if it reads as a typo it's the wrong word choice. Spelled out as "if and only if" in the amended section.

docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md
bdarcus: Two parts to this. First, a clarification: `render-when` itself predates this PR — `TemplateGroupCondition` with `field-present`/`field-absent` is shipped schema, used in e.g. chicago-author-date-18th.yaml. This spec only proposed adding two condition kinds to it.

That said, the pushback is right by our own doctrine: the 2026-07-12 schema-issue triage explicitly declines template-conditional growth as a category (#62/#320 — value-testing conditionals get solved in options/presets, not template syntax), and which role is redundant when two roles coincide is a property of the style, not of a template position.

Reworked (amended into d0eefa7): the elision case is now an options rule —

```yaml
contributors:
  suppress:
    - role: performer
      when-identical-to: composer
```

— named `suppress` per your suggestion, consistent with `Rendering.suppress`/`suppress-author`. When the rule fires (person sets non-empty and identical), the suppressed role renders empty everywhere and existing group empty-collapse removes the `[Recorded by …]` descriptor. Engine-side it slots into the same path as substitute-driven role suppression (`is_role_suppressed_by_substitute`). `render-when` is untouched by the spec.

bdarcus: Follow-up on the broader "why do we need this in general" question: usage census shows render-when is confined to the two Chicago 18th embedded styles (29 sites; migrate never emits it), so removal is plausibly within reach. Tracked as bean csl26-qyub (audit: map each usage to an options/preset/type-variant replacement, then decide remove vs freeze), committed on this branch in df5ff3e. It also blocks csl26-q05f, which would otherwise grow the mechanism.

PR #1052, opened 2026-07-13
Cross-Role Contributor Lists

bdarcus:
## Summary

- add explicit multi-role contributor data plus scalar/list style schema support, validation, locale terms, and generated schemas
- render ordered cross-role lists with individual or collective labels, authored/composed combined labels, post-combination et-al, and exact explicit-role suppression
- use merged contributors for sorting and disambiguation, preserve multi-variable CSL names during migration, and update APA multimedia behavior and fixtures

Native Citum YAML is authoritative: a contributor explicitly carries `roles: [...]`. The engine never infers identity from display names. Strict trimmed NFC identity matching exists only at the legacy CSL-JSON conversion boundary, where independent role arrays must be normalized into native contributor entries.

## Review hardening

A full spec/schema/engine/migration review pass folded the following fixes into this PR:

- merged lists now apply `role-substitute` suppression, so a role rendered as a primary-slot fallback is excluded from merged assembly (spec: Interactions)
- bibliography contributor matching (subsequent-author substitution) now uses the shared effective-primary resolver, honoring type overrides and merged candidates
- `role.omit` now also suppresses style-wide structural role labels (`role.roles.<role>.label`, `role.defaults` bundles) on both scalar and merged paths, with the existing verb-form exception
- empty-name substitute candidates (e.g. `editor: []`) no longer terminate the fallback chain; the next candidate is tried, restoring title fallback in rendering and disambiguation keys
- an empty merged primary component now sorts through the same substitute chain the render path uses instead of jumping to the title key
- grouped author-date citations recognize merged `[author, …]` leading components; migration gating and SQI compaction handle list-form components
- migration emits a per-role label override for every role in a multi-variable `<names>` block (previously only the first role was labeled, dropping e.g. the translator label when editor and translator differ)
- removed invented `writer`/`writer-director` short and verb-short locale abbreviations; short-form lookups fall back to the long forms
- spec: documented empty-merged-component fallback, component `merge:` wholesale replacement of style-wide defaults, and editor-translator combined-term order-insensitivity

Sort/disambiguation paths now resolve the effective primary once per reference instead of per comparison (single-pass resolver, precomputed sort keys, borrow-based merge config — `csl26-78ds`): rendering benchmarks improved 14–47% versus the prior baseline. Deferred to a bean: validation/label-type consolidation (`csl26-tsmg`).

## Validation

- `just pre-commit` — 1,986 tests passed; formatting and clippy clean
- `just schema-gen` — no drift against checked-in schemas
- APA oracle — citations 20/20; bibliography 45/46, with only the pre-existing standard-reference mismatch remaining
- new regression tests cover every review fix (each verified red-before/green-after)

## Tracking

- completes and archives bean `csl26-7ip9`; review tracked in `csl26-nnkf`; sort-path performance in `csl26-78ds`
- activates `docs/specs/CROSS_ROLE_CONTRIBUTOR_LISTS.md` (v1.0) and `docs/specs/PRIMARY_CONTRIBUTOR_SUBSTITUTION.md` (v1.0)
- follow-ups: `csl26-tsmg`, `csl26-h6b9`, `csl26-hyc4`, `csl26-ijvf`

### DIFF
--- a/crates/citum-engine/src/sorting.rs
+++ b/crates/citum-engine/src/sorting.rs
@@ -25,9 +25,14 @@ use citum_schema::reference::ClassExtension;
 
 use crate::reference::Reference;
 use crate::sort_support::{
-    SortKeyOptions, TextCollator, author_sort_key_opt_with_options, collator_locale_id,
-    normalize_sort_text, title_sort_key_with_options,
+    SortKeyOptions, TextCollator, collator_locale_id, flat_names_sort_key, normalize_sort_text,
+    title_sort_key_with_options,
 };
+
+enum PrimaryContributorSpec<'a> {
+    Citation(&'a citum_schema::CitationSpec),
+    Bibliography(&'a citum_schema::BibliographySpec),
+}
 
 fn compare_optional_years(a_year: Option<i32>, b_year: Option<i32>) -> std::cmp::Ordering {
     match (a_year, b_year) {
@@ -43,6 +48,9 @@ pub struct ReferenceSorter<'a> {
     locale: &'a Locale,
     text_collator: TextCollator,
     sort_key_options: SortKeyOptions,
+    config: Option<&'a Config>,
+    primary_contributor_spec: Option<PrimaryContributorSpec<'a>>,
+    primary_contributor_may_be_list: bool,
 }
 
 struct CachedReference<'a> {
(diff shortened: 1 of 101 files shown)


crates/citum-schema-style/embedded/locales/en-US.yaml
```diff
@@ -245,6 +245,24 @@ roles:
...
+    verb-short: writ. by
+  writer-director:
+    long:
+      singular: writer & director
+      plural: writers & directors
+    short:
+      singular: wrtr. & dir.
+      plural: wrtrs. & dirs.
+    verb: written & directed by
+    verb-short: writ. & dir. by
```
bdarcus: but the conjunction needs to be configurable? Does this not suggest we should be using MF2 here?

crates/citum-schema-style/embedded/styles/apa-7th.yaml
```diff
@@ -621,10 +639,18 @@ bibliography:
...
+    - contributor: [writer, director]
       form: long
       name-order: family-first
       suffix: "."
+      merge:
+        order: document
+        labels: individual
+        roles:
+          writer:
+            label: {term: writer, form: long, placement: suffix, text-case: title, prefix: " (", suffix: ")"}
```
bdarcus: this should use `wrap`; no?

crates/citum-schema-style/embedded/styles/apa-7th.yaml
```diff
@@ -621,10 +639,18 @@ bibliography:
...
+          director:
+            label: {term: director, form: long, placement: suffix, text-case: title, prefix: " (", suffix: ")"}
```
bdarcus: this should use wrap; no?

crates/citum-engine/src/sorting.rs
```diff
@@ -411,6 +455,21 @@ impl<'a> ReferenceSorter<'a> {
         )
     }
 
+    fn primary_contributor_component(
```
bdarcus, line 555: **Performance (sort/disambiguation hot path):** `primary_contributor_component` materializes a full owned `Template` (`resolve_template_for_type` → `variant.clone().into_template()`) **and** clones a `TemplateContributor` (`first_contributor_component`) on *every* call — purely to check `component.contributor.is_multiple()`. For every style shipped today (none use list-form contributors) this always concludes "not multiple → old path," so the clone is pure waste.

It's invoked once per reference in the cached path (`cache_sort_value` → `extract_author_sort_key_opt`, O(n) template clones over the whole bibliography) and per-comparison in the `compare_by_key` `sort_by` loops at `setup.rs:625` (small n) and `disambiguation.rs:791` (group-bounded).

Suggest gating the work on whether the effective spec actually contains a multiple-role contributor (short-circuit before resolving), or resolving the primary component once per ref-type rather than per reference. Flagging for the author — not auto-fixing, since it's a hot-path design decision.

crates/citum-engine/src/values/contributor/names.rs
```diff
@@ -432,6 +457,68 @@ pub fn format_names(
...
+                format_single_name(name, form, original_index, context, expand),
+                original_index,
+                decorations,
+            )
+        })
+        .collect();
+    (formatted_first, formatted_last)
+}
+
+fn decorate_name(value: String, index: usize, decorations: &[NameDecoration]) -> String {
```
bdarcus, line 516: **Efficiency (name-formatting path):** `decorate_name` uses `map_or(value.clone(), …)`, whose eager default clones the formatted name on **every** call — including the decorated branch, where the clone is then discarded. Undecorated names (the common case) also pay a clone.

Applied fix in the working tree: `match` returning `value` by move on the no-decoration arm, and skipping the `format!` allocation entirely when both affixes are empty.

```rust
fn decorate_name(value: String, index: usize, decorations: &[NameDecoration]) -> String {
    match decorations.get(index) {
        Some(decoration) if !decoration.prefix.is_empty() || !decoration.suffix.is_empty() => {
            format!("{}{value}{}", decoration.prefix, decoration.suffix)
        }
        _ => value,
    }
}
```

PR #1068, opened 2026-07-18
add date annotation as a note

bdarcus:
## Summary

Implements `docs/specs/CALENDAR_DATE_ANNOTATIONS.md`: any EDTF date value
may carry an opaque `note` (e.g. a source-calendar annotation like
`1947（民国三十六年）`) alongside its canonical `value`, rendered only when
a style opts in. Spec status: **Active**.

## Design

**Data model.** Every date field accepts either the existing EDTF scalar or
a structured `{ value, note }` mapping. `value` is the canonical EDTF
value; `note` is opaque text Citum never parses, converts, or validates.
Scalar input/output shapes are unchanged when no note is present.

**Type change — extend in place, no new wrapper.** The canonical date type
(`EdtfString`) is renamed to `DateValue` and gains `note: Option<String>`,
rather than introducing a parallel wrapper type around it. This changes
**zero** field declarations across the ~20 reference structs that use
it (`issued: EdtfString` → `issued: DateValue` is a token rename, not a
per-field reshape); custom scalar/mapping serde and a matching `oneOf` JSON
schema live once on the type itself. Value-oriented accessors
(`parse()`/`year()`/`is_range()`/…) and all processing (sorting,
disambiguation, collision grouping, fallback selection) read only
`.value`, never `.note`.

**Render opt-in.** A style enables notes through `note-wrap` in its
**date** configuration, reusing the existing `wrap` value model:

```yaml
bibliography:
  options:
    dates:
      note-wrap: parentheses
```

- **Bibliography-only scoping for free.** `DateConfig` is embedded at the
  global, citation, and bibliography scopes, matching its existing
  `era-labels`/`range-delimiter`/`approximation-marker` siblings. Setting
  the option under the bibliography block scopes note rendering there;
  citations render no note with no separate toggle.
- **Script-appropriate delimiters** come from the existing `csl26-k2kp`
  script-aware wrap renderer (`realize_wrap`) — full-width `（ ）` for
  CJK-script items, half-width `( )` for Latin — with no new normalization
  pass added by this feature.
- The note trails the complete formatted date, after any inlined
  year-suffix disambiguator (`1947a（民国三十六年）`).

**Legacy conversion.** The CSL/Zotero-extra convention
`issued: 1947（民国三十六年）` (full-width parentheses only, disjoint from
the `c1988`/`1995印刷`/`1936~` publication-year substitutes) converts to
`{ value, note }` in `citum-schema-data`'s reference conversion
(`annotated_issued_from_legacy`), alongside the existing
copyright/printing/estimated-year handling.

**Style enablement.** `note-wrap: parentheses` is set in the GB/T
7714—2025 author-date style's bibliography `dates` options.

## Revision history

- **v1.0–v1.1:** initial draft; review found the original per-`TemplateDate`
  `calendar-note: parentheses` opt-in unbuildable (no full-width wrap path
  existed yet). Redesigned around the reused `wrap` vocabulary, gated on a
  not-yet-built script-aware wrap renderer.
- **v1.2:** retargeted that prerequisite to increment 1 of the punctuation
  realization layer (bean `csl26-k2kp`).
- **v1.3:** generalized `calendar-note` → `note` (opaque, non-processed
  text — reserves the `calendar`/`era`/`original` namespace for a future
  genuinely *computed* regnal-year feature, bean `csl26-0kqf`) and moved
  the opt-in from `MultilingualConfig.calendar-note-wrap` to a flat
  `DateConfig.note-wrap`, once the script-aware wrap prerequisite landed.
- **v1.4 (this PR's implementation):** ships the feature described above;
  flips spec Status Draft → Active.

## Tests

15 new tests: `DateValue` serde round-trip (scalar byte-identical,
mapping parse, unknown-field rejection), HTML-escaping of the note,
script-aware wrap for a single date and an interval, year-suffix
ordering, author-date collision-invariance (two same-author-same-year
references with different notes still collide, verified end-to-end
through real disambiguation logic), bibliography-vs-citation scoping
against the real embedded GB/T style, and legacy-conversion tests
including a half-width-parenthesis disjointness check.

## GB/T author-date audit (unchanged from prior revisions)

- All eight existing author-date citation scenarios pass (pre-existing
  suite; not independently re-run against this change via the
  oracle/workflow-test harness — see Known gaps).
- The two §7.5.4.1 records render only bare Gregorian years in both
  citeproc-js and current Citum; the broader standard-derived benchmark
  remains tracked by `csl26-9qat`, and the general author-date bibliography
  tune by `csl26-6eak`.

## Validation

- `just pre-commit` (fmt --check + clippy -D warnings + full nextest,
  2094 tests) — green.
- `just schema-gen` — regenerated schemas committed.
- All CI checks pass: Rust CI, Fidelity Checks, API Semver Check, Hygiene
  Checks (Beans/Docs/Infra), Security Audit, WASM Benchmark Smoke Test,
  Release Dry Runs.

## Known gaps (tracked, not silently skipped)

- The eight GB/T author-date citation scenarios have not been re-verified
  through the oracle/workflow-test harness against this specific change
  (the pinned-corpus gap noted in an earlier revision is resolved — see
  below).

Tracked in bean `csl26-epu6` (completed) and `csl26-0kqf` (the separate,
still-unimplemented computed regnal-year feature this data model
unblocks).

## Follow-up from review

- **Rolled out to all three GB/T styles**, not just `-author-date`:
  `note-wrap` is now also set on `gb-t-7714-2025-numeric` and
  `gb-t-7714-2025-note`. Verified against the real pinned corpus
  (`tests/fixtures/test-items-library/gb-t-7714-2025.json`), which turned
  out to already contain **five** annotated records, not just the two
  §7.5.4.1 examples: `gbt7714.7.5.4.1:1`/`:2` (Minguo/Kangxi), plus
  `gbt7714.8.2.2:2` (Qing Tongzhi era), `gbt7714.8.12.3:1` (Guangxu era),
  and `gbt7714.8.12.3:3` (Republic era). Added a Rust regression test
  against the real corpus (`crates/citum-engine/tests/date_annotations.rs`,
  9 cases) and registered the expected citeproc-js divergence
  (`div-015`, `docs/adjudication/DIVERGENCE_REGISTER.md`). `gb-t-7714-2025-numeric`
  turned out not to be in `core-quality-baseline.json`'s CI-gated style
  set, so no oracle-comparator masking code was needed.
- **Under discussion, not decided or implemented:** whether `note` should
  serve as a literal display fallback when a date has no computable EDTF
  `value` at all — i.e., a date entered as pure calendar-note prose
  renders via `note` instead of falling through to the locale's "n.d."
  term. Raised in review
  ([discussion](https://github.com/citum/citum-core/pull/1068#discussion_r3616022626)).
  Related to, but distinct from, the already-documented "bare (unwrapped)
  note" render extension point in
  `docs/specs/CALENDAR_DATE_ANNOTATIONS.md` — that one is about
  delimiter-free *rendering* of an existing note; this is about *input*
  with no `value` at all. Today a `DateValue` with an empty `value` is
  filtered out before rendering ever reaches the note-wrap step
  (`date_opt.filter(|d| !d.is_empty())`,
  `crates/citum-engine/src/values/date.rs`), so the note would currently
  be silently dropped rather than shown. Not scoped for this PR.


docs/specs/CALENDAR_DATE_ANNOTATIONS.md
```diff
@@ -0,0 +1,291 @@
...
+
+```yaml
+bibliography:
+  options:
+    dates:
+      note-wrap: parentheses
+```
+
+`DateConfig.note_wrap` (`crates/citum-schema-style/src/options/dates.rs`) is
+an optional `WrapConfig`, a flat field alongside `DateConfig`'s other
```
bdarcus, line 133: A problem with this approach is it doesn't allow a bare note. It's possible that won't ever be needed, but we need to anticipate, at least in a doc comment, that it might be in the future?

bdarcus: Documented as an extension point in v1.5 (commit e4193621) — note-wrap always applies one of parentheses/brackets/quotes today; a bare/unwrapped render would need a WrapPunctuation::None variant on the shared wrap vocabulary. Not implemented, since no known style needs it yet.

bdarcus: We also likely want to consider using `note` as a literal fallback, if there is no (EDTF) value?

PR #1083, opened 2026-07-22
feat(locale): add localized digit systems

bdarcus:
## Summary

Adds locale-configured digit rendering for `number` template components, with Arabic-Indic digits enabled for the embedded `ar-AR` locale.

## Scope

- Adds the public `number-formats.digit-system` schema enum and regenerated locale schema.
- Localizes ASCII digits after ordinal and label resolution while preserving letters and punctuation.
- Documents the locale contract and completes bean `csl26-5q59`.

## Validation

- `just schema-gen`
- `just pre-commit`
- Pre-push hook: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo nextest run`

## Risk

Existing locales default to Western digits. Coverage includes the default, Arabic-Indic, extended Arabic-Indic, Devanagari, ranges, and mixed identifiers.

### DIFF
--- a/crates/citum-engine/src/values/number.rs
+++ b/crates/citum-engine/src/values/number.rs
@@ -10,7 +10,7 @@ SPDX-FileCopyrightText: © 2023-2026 Bruce D'Arcus and Citum contributors
 
 use crate::reference::Reference;
 use crate::values::{ComponentValues, ProcHints, ProcValues, RenderOptions};
-use citum_schema::locale::{GeneralTerm, GrammaticalGender, MessageArgs, TermForm};
+use citum_schema::locale::{DigitSystem, GeneralTerm, GrammaticalGender, MessageArgs, TermForm};
 use citum_schema::reference::ClassExtension;
 use citum_schema::template::{LabelForm, NumberForm, NumberVariable, TemplateNumber};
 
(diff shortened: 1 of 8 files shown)


crates/citum-engine/src/values/number.rs
```diff
@@ -377,6 +402,38 @@ mod is_numeric_tests {
     }
 }
 
+#[cfg(test)]
+mod digit_system_tests {
+    use super::localize_digits;
+    use citum_schema::locale::DigitSystem;
+
+    #[test]
+    fn given_digit_system_when_localizing_then_only_ascii_digits_change() {
```
bdarcus: Don't project rules explicitly say NOT to use BDD-style test naming for non-(multi-parameter)BDD tests? If this IS a BDD test, use `rstest`?

crates/citum-engine/src/values/number.rs
```diff
@@ -377,6 +402,38 @@ mod is_numeric_tests {
...
+        ] {
+            assert_eq!(
+                localize_digits("AB-12, 34".to_string(), &digit_system),
+                expected
+            );
+        }
+    }
+
+    #[test]
+    fn given_unknown_digit_system_when_localizing_then_western_digits_are_preserved() {
```
bdarcus: same issue; test name?

PR #1087, opened 2026-07-23
fix(engine): disamb grouping and anon term

bdarcus:
## Summary

Continues csl26-6eak (Tune `gb-t-7714-2025-author-date` to full fidelity). Root-causes and fixes three independent engine gaps blocking year-suffix disambiguation for GB/T's anonymous-author entries — found via direct instrumentation (`ProcHints` dumps, CLI render diffs) at each step rather than assumption:

1. **Disambiguation grouping used the citation-scoped substitute, not the bibliography-scoped one** (`disambiguation.rs`) — a style overriding only the bibliography-scope substitute never collided same-year anonymous entries for suffix assignment.
2. **`TemplateDate`'s explicit `fallback:` branch never computed or attached a disambiguation suffix** (`date.rs`) — the suffix mechanism only existed in the implicit no-date and real-date branches; GB/T author-date's date components always use an explicit `fallback:` chain.
3. **The anonymous term (`佚名`/`Anon`) was a locale-independent style constant**, and grouping wasn't scoped by item language either — `general_message_id`'s allowlist omitted `Anonymous` (unlike `NoDate`), so a locale's own message catalog was never consulted; fixed to mirror `no-date`'s existing per-locale mechanism exactly, and scoped the grouping sentinel by item language so Chinese/English anonymous entries get independent letter sequences (matching oracle).

Three previously-passing pinned tests regressed once (3) landed — fixed the test harness (construct with the embedded zh-CN locale, matching how the real `citum render` CLI resolves a style's `default-locale`), not the assertions.

**Fidelity**: raw 125/203 → 141/203 (69.5%), adjusted 152/203 → 162/203 (79.8%) against the upstream GB/T corpus. Not yet at the 100% bar `verification-policy.yaml` requires, so it's left untouched.

Residue is precisely characterized and split into follow-up beans:
- csl26-m8la (highest leverage, ~28 of 41 remaining failures): `sort_group_for_year_suffix`'s no-`group_sort` fallback hardcodes a title-alphabetical tiebreak that doesn't match this style's actual render order — a shared path needing its own dedicated corpus-verified pass.
- csl26-yyrs: org-as-author promotion for the `standard` type (3 entries).
- csl26-o333: approximate-year/dropped-body-date edge cases (5 entries).

## Test plan

- [x] `just pre-commit` (fmt, clippy, 2161/2161 nextest) — green
- [x] `just check-core-quality` (157 embedded styles, fidelity=1.0, 0 warnings) — green, no regression from the shared `en-US.yaml` term.anonymous default change
- [x] GB/T author-date oracle corpus (bibliography scope, `tests/fixtures/test-items-library/gb-t-7714-2025.json`) — 141/203 raw, 162/203 adjusted, up from 125/152
- [x] `date_annotations.rs` (13/13) — including the 3 harness-fixed pinned era-annotation cases

### DIFF
--- a/crates/citum-schema-style/embedded/locales/en-US.yaml
+++ b/crates/citum-schema-style/embedded/locales/en-US.yaml
@@ -933,6 +933,10 @@ messages:
   term.retrieved: "retrieved"
   term.no-date: "n.d."
   term.no-date-long: "no date"
+  # GB/T 7714's own anonymous-author convention (not the CSL legacy
+  # `terms.anonymous` "anon." catalog entry) — see csl26-6eak.
+  term.anonymous: "Anon"
+  term.anonymous-long: "Anonymous"
   term.forthcoming: "forthcoming"
   # CSL reference locale (scripts/locales-en-US.xml): CMOS 10.48 allows
   # "ca." but CSL has historically used "c." — match the CSL reference.
--- a/.beans/csl26-6eak--tune-gb-t-7714-2025-author-date-to-full-fidelity.md
+++ b/.beans/csl26-6eak--tune-gb-t-7714-2025-author-date-to-full-fidelity.md
@@ -1,7 +1,7 @@
 ---
 # csl26-6eak
 title: Tune gb-t-7714-2025-author-date to full fidelity
-status: todo
+status: in-progress
 type: task
 priority: normal
 tags:
@@ -9,7 +9,7 @@ tags:
     - fidelity
     - multilingual
 created_at: 2026-07-16T10:56:59Z
-updated_at: 2026-07-23T15:05:20Z
+updated_at: 2026-07-23T17:05:29Z
 blocking:
     - csl26-dxuo
 blocked_by:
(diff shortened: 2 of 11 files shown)


crates/citum-schema-style/embedded/locales/en-US.yaml
```diff
@@ -933,6 +933,10 @@ messages:
   term.retrieved: "retrieved"
   term.no-date: "n.d."
   term.no-date-long: "no date"
+  # GB/T 7714's own anonymous-author convention (not the CSL legacy
+  # `terms.anonymous` "anon." catalog entry) — see csl26-6eak.
+  term.anonymous: "Anon"
+  term.anonymous-long: "Anon"
```
bdarcus: Why wouldn't this be "Anonymous"?

bdarcus: Good catch — should have followed the term.no-date/term.no-date-long convention right above it. Fixed: term.anonymous-long is now "Anonymous" (matching the pattern of "n.d."/"no date"). zh-CN.yaml's version is correct as-is — its own no-date-long is also identical to no-date, so 佚名/佚名 was already consistent with the established local pattern.

PR #1103, opened 2026-07-26
feat(engine): configure EDTF year ranges

bdarcus:
## Summary

- Add `options.dates.range-format` with `expanded` (default) and `chicago` modes, independent of page-range formatting.
- Reuse Chicago inclusive-number abbreviation for closed, year-only same-era CE and BCE intervals; cross-era, reversed, open, and unspecified ranges retain endpoint rendering.
- Add MF2 `pattern.date-range-<form>` messages for shared-year intervals, with Spanish examples such as `mayo a junio, 2026`.
- Enable Chicago 18 in the shared base and document the configuration and locale-message contract.

## Validation

- `just schema-gen`
- Focused EDTF range suite: 18 passing tests
- Chicago oracle: 20/20 citations and 46/46 bibliography
- Batch workflow check: Chicago remains 20/20 citations and 46/46 bibliography
- `just pre-commit` (fmt, clippy, 2210 nextest cases)

## Risk

`expanded` remains the default. Chicago abbreviation is limited to fully specified, closed, year-only intervals in one era. Shared-year locale messages are opt-in; locales without a pattern retain existing fallback layouts. Distinct-year grammatical interval patterns remain a follow-up.

### DIFF
--- a/crates/citum-engine/src/values/date.rs
+++ b/crates/citum-engine/src/values/date.rs
@@ -12,7 +12,7 @@ use crate::reference::{DateValue, Reference};
 use crate::values::{ComponentValues, ProcHints, ProcValues, RenderOptions};
 use citum_edtf::{Edtf, Timezone, UnspecifiedYear, Year};
 use citum_schema::locale::{GeneralTerm, TermForm};
-use citum_schema::options::dates::TimeFormat;
+use citum_schema::options::dates::{DateRangeFormat, TimeFormat};
 use citum_schema::reference::types::RefDate;
 use citum_schema::reference::{ClassExtension, WorkRelation};
 use citum_schema::template::{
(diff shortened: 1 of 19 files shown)


crates/citum-engine/src/values/date.rs
```diff
@@ -1742,6 +1775,34 @@ mod range_tests {
...
+            chicago_range(&en_us(), "2021/2026", DateForm::Year).as_deref(),
+            Some("2021–26")
+        );
+    }
+
+    #[test]
+    fn chicago_range_format_does_not_condense_non_year_forms() {
+        assert_eq!(
+            chicago_range(&en_us(), "2021-05/2026-06", DateForm::YearMonth).as_deref(),
+            Some("May 2021–June 2026")
```
bdarcus, line 1939: While I can't see the Chicago guide addressing this case, shouldn't expected here be "May–June 2026"?

EDIT: may bad; read too quickly.

We may need to leave room to expand this into MF2 messages in the locale system though.

PR #1142, opened 2026-08-05
docs(spec): template layering and all precedence

bdarcus:
**Docs only — the review gate before any schema code.** No Rust, no `schema-gen`, no style changes.

> **Rewritten after review.** The earlier draft proposed three schema additions. Your comment — that title quoting isn't type-specific and belongs in the title preset — was right, and checking it **withdrew two of the three**. What's left is the layering rule they were working around, plus a live bug found while checking the scoping.

## What the review found

`ieee.yaml` set `titles: humanities`, a preset whose `component` rendering is **plain**. IEEE quotes article titles, so the style compensated with `wrap: { punctuation: quotes }` on the **base** `title: primary` — applying it to every reference type — then pasted six near-identical type-variants whose only job was to take the quotes back off. One of them was **byte-identical to the base template**.

`TitlePreset::Ieee` already exists and quotes component titles. The style was using the wrong one, and `unset` was a proposal to make the workaround more comfortable.

**Proven, not argued:** #1143 moves the policy to `options.titles`. **361 → 233 lines, ieee exact parity 95/149 unchanged, full embedded sweep 1558 → 1559, zero regressions.** With the wrap gone, the remaining variants differ from the base only by *additions*, which `modify` already expresses.

Across all 32 embedded styles: 155 full-replacement type-variants, 148 structurally different, and every field ever needing to be *cleared* was `title.wrap` — 5 occurrences, all ieee, all this same workaround. **Post-fix demand for `unset` is zero.** `abstract-variants` was only ever a readability win over multi-type keys, which already work.

## What the spec now says

**§5.1 — templates own structure, options own rendering policy.** The boundary isn't stylistic: category config is read by **paths a template cannot reach**. When a title substitutes for a missing author, `resolve_author_substitute` deliberately passes no component rendering into the title branch — its own doc comment says so — so quoting resolves from `options.titles` alone. A template `wrap:` gets substituted titles wrong no matter how many variants you write. That is your original point, and it's now the stated rule.

**§5.2 — the `all` selector, with specificity-based precedence.** This is the defensible core of "these rules aren't type-specific": `all` already exists and matches every type, so a generic diff *is* expressible. But precedence is currently first-match-in-**authored order**, with no specificity ordering anywhere in the engine. The spec now requires exact type > multi-type key > `all`.

**§5.3 — presets are not macros.** Your aside, assessed: the analogy doesn't hold. A macro is a named fragment of *template*, invoked from a template body, composable. A preset is a named bundle of *policy*, referenceable only from an `options.*` slot, non-composable. Presets are the concrete form of §2's "Logic-Heavy Options" — they absorb what macros were *used for* without adopting what macros *are*. Acting on the analogy leads back to exactly the structure ieee got wrong.

**§5.4** restates the prohibition, now that §5.2 introduces a selector spanning types.

**G3 (`LocaleOverride.vocab`) is unchanged** and independently motivated by IEEE's `Ph.D. dissertation`.

## Live bug found while checking this

`elsevier-with-titles-core.yaml` is the only embedded style using `all`. Its own comment states the rule correctly — *"any type-specific variant must precede it to take effect"* — and the file then **violates it**: `all` sits at line 109 with **16 type-variant keys authored after it**, all unreachable. Confirmed from rendered output: ITEM-14 is a book but renders with the `all` variant's `(Eds.)` editor-label shape, not the `book` variant's. The style sits at **27/67**.

Filed as `csl26-q4g5`. Fixing it makes 16 variants live at once and needs its own PR and full sweep — blocked on the precedence decision in §5.2, which is what this PR is asking you to make.

## Related

- #1143 — the ieee change that proves the layering argument. Worth reviewing alongside this.
- `csl26-8u5i` — `Style.templates` ("Named reusable templates") is declared, validated, merged, and never resolved by any reference form.

**Open question I'd like your call on:** `titles.type-mapping` is `HashMap<String, String>`, validated nowhere, looked up with a plain `get` against `Reference::ref_type()`. An underscore spelling that works as a `type-variants` key silently never matches, and an invalid category value silently falls through to `default`. Both vocabularies already exist as types — `validate_type_name` (which already normalizes) and the `TitleCategory` enum. The field just bypasses them.


docs/specs/TEMPLATE_V3.md
```diff
@@ -302,19 +302,229 @@ bibliography:
...
+
+#### 5.1 Clearing an inherited field (`unset`)
+
+A `modify` operation can set a rendering field but not clear one. `Rendering`
+fields are all `Option`, and the merge assigns only where the incoming value is
+`Some`, so there is no value an author can write that means "the parent set
+this; I don't want it."
+
+The practical consequence is severe. A base template whose `title: primary`
+carries `wrap: { punctuation: quotes }` cannot be specialized into a variant
```
bdarcus: Limiting this to type-variants seems unwise, given that I don't think rules like are type-specific? If, for example, a primary use case is a citation where a title is substituting for a missing author, that would be a generic rule.

Aside: I've been loosely thinking that Citum presets are something like typed and more constrained CSL macros. Perhaps we should assess that idea and articulate it it holds?

PR #1150, opened 2026-08-06
fix(engine): year-suffix follows render order (csl26-m8la)

bdarcus:
## Summary

- **Bean's original diagnosis was wrong.** csl26-m8la blamed `Disambiguator::sort_group_for_year_suffix`'s *no-`group_sort`* fallback (title-alphabetical tiebreak). That branch is never reached by the reported style, `gb-t-7714-2025-author-date` — it always resolves a `group_sort` (inherited `citation-number` from its numeric base). The real defect is in the *other* branch: when a resolved sort's keys don't fully order a collision group, the pre-sort fell back to title order instead of the renderer's actual fallback (registry order, or id order when set).
- **Fixed**: `sort_group_for_year_suffix` now mirrors `ReferenceSorter`'s real fallback chain — empty template → registry order untouched; non-empty template → id (if set) → full issued date → registry index.
- **A second defect was found and deliberately left unshipped.** `gb-t-7714-2025-author-date.yaml` is still missing its own `bibliography.sort` (still inherits `citation-number` from its numeric base — registry order, not a real author+date order). Restoring it, plus a companion `sorting.rs` Author-key fix needed to keep that restoration net-neutral, measured out **net-negative**: no improvement on this style's own oracle beyond the engine fix alone, and a severe regression elsewhere (`american-medical-association-alphabetical` exact-parity 21/67 → 1/67) via a shared code path touching all 157 styles. Filed as `csl26-q67h` with full measurements, not bundled here.

## Numbers

`gb-t-7714-2025-author-date` bibliography oracle (`--refs-fixture gb-t-7714-2025.json`): adjusted failures **42 → 30** (of 203; 173 passed). Not full oracle parity — the style still renders in registry order — but internally consistent with citum's own render, which is the failure mode the bean reported. A residual ~9-entry gap traces to a separate architectural mismatch (variable-based vs. render-text-based collision grouping), tracked in `csl26-huuz`.

Full 35-style exemplar corpus vs. `main`: **4 styles improved** (gb-t-7714-2025-author-date, elsevier-harvard, springer-basic-author-date, taylor-and-francis-cse-author-date), **1 single-entry regression** (`entomological-society-of-america`, 9/67 → 8/67 — root cause not isolated within a bounded investigation, tracked in `csl26-92kv`).

Zero regressions in `citum-engine`'s test suite (2420 workspace tests pass, including 19 new/updated unit tests and 2 new integration tests for this fix). `just pre-commit` and the pre-push gate both clean.

## Also in this PR

- `csl26-0h05` ("Type titles.type-mapping keys and categories") is stale — all six acceptance criteria landed in `005f0741`/`b996fb69`/`613ffb94`/`81555bba`/`b460bc6f`. Marked completed and archived.
- Three new follow-up beans filed: `csl26-q67h` (restore gb-t's own sort — the reverted defect above), `csl26-a19q` (`citation_number_sort_not_supported` warning misses inherited `bibliography.sort`), `csl26-92kv` (the ESA residual regression).
- `csl26-huuz` (pre-existing architectural finding: disambiguation collision grouping is variable-based, not render-text-based) is also included, wording updated to reflect the final shipped scope.

## Open questions for review

- Should `csl26-q67h` (restoring gb-t's own bibliography sort) be prioritized, given it's the difference between "internally consistent" and "actually correct" for this style?
- The `entomological-society-of-america` regression (`csl26-92kv`) is shipped as a known, documented residual per an explicit accept-and-move-on call during review — flagging in case that tradeoff should be revisited before merge rather than after.

## Test plan

- [x] `cargo nextest run` (workspace, via pre-push gate): 2420/2420 passed
- [x] `cargo fmt --check` / `cargo clippy --all-targets --all-features -- -D warnings`: clean
- [x] Targeted oracle: gb-t-7714-2025-author-date bibliography, 173/203 (30 failed), matching pre-regression baseline
- [x] Full 35-style exemplar corpus (`report-core.js --all-features --parallelism 1` + `check-core-quality.js`): gate passed, diffed against clean `main` baseline
- [ ] `gh pr checks` — pending


crates/citum-engine/src/processor/disambiguation.rs
```diff
@@ -798,18 +825,59 @@ impl<'a> Disambiguator<'a> {
...
+            // comparison below may run for it; registry order must survive untouched.
+            // Only once the template is non-empty does the renderer's stable sort
+            // actually engage those finer tiebreaks, so both steps are gated on that,
+            // not just on `id_tiebreak`.
+            //
+            // A resolved `Issued` sort key only compares the YEAR
+            // (`ReferenceSorter::issued_year`), so two same-author/same-year references
+            // that differ by month/day still tie under the real template even though
+            // `cached.data.index` alone doesn't reliably predict which one the renderer
+            // puts first — confirmed by direct comparison against
```
bdarcus: Is this not a bug/design flaw? If there are multiple references from the same year, let's say from a news reporter, then those year references should sort by the full ISO/EDTF date; e.g. Jan 1 comes first (2020a), December 31 last (2020g).

crates/citum-engine/src/processor/disambiguation.rs
```diff
@@ -798,18 +825,59 @@ impl<'a> Disambiguator<'a> {
...
+            // `chicago-author-date-18th`'s rendered bibliography for two same-year,
+            // no-title Gourmet magazine entries (May vs. September): registry-index-only
+            // put September first, but the actual render puts May first. Comparing full
+            // issued dates (`year_suffix_date_key`) before falling back to `index`
+            // matches the render for every case checked so far, but the underlying
+            // mechanism isn't fully understood — `chicago-author-date-18th` has a
```
bdarcus: "isn't fully understood"?? Seems we should endeavor to fully understand it before merging this PR.

crates/citum-engine/src/sorting.rs
```diff
@@ -604,11 +607,20 @@ impl<'a> ReferenceSorter<'a> {
...
+    /// Full (year, month, day) for sort comparison, not just the year.
+    ///
+    /// Two same-year references with different months previously tied under
+    /// an `Issued` sort key, which only ever compared the year - so
+    /// `sort_by_keys` fell through to its next key or a stable-sort tie,
+    /// rather than the chronological order most citation styles expect
+    /// (Jan before Dec within the same year). Missing month/day default to 0
+    /// (sorts before any real month/day), matching EDTF's own precision
+    /// convention that a less-precise date could be anywhere in its range.
+    /// csl26-m8la, PR #1150 review.
```
bdarcus: I don't want docs to refer to what we previously did (a bug); it's not helpful. Rewrite this to state what this function now does.

PR #1171, opened 2026-08-12
fix(engine): grouping follows resolved date slot

bdarcus:
**Fixes csl26-huuz** — year-suffix collision grouping now follows what a
reference's resolved date slot actually renders, instead of assuming every
undated reference renders the same "no date" text. This is Layer 1 of the
date-substitute stack: the candidate-neutral engine foundation on which the
Draft spec and implementation layers build.

- `disambiguation.rs`: the date-slot discriminant reads the resolved template
  when no issued year is parseable. The selected identity slot is paired with
  its effective scope configuration: bibliography options for a bibliography
  slot, citation options for the citation fallback slot.
- `values/date.rs`: fallback discrimination accepts date form, rendering, and
  `suppress-note` independently, so the implementation layer can supply either
  inline or options-level candidates without duplicating the algorithm.
- Fallback year-suffix letters land inside candidate wrapping (`[2020a]`, not
  `[2020]a`), while an empty fallback chain still renders the letter standalone.
- `gb-t-7714-2025-author-date.yaml`: two type variants' date fallbacks match
  the upstream CSL-M macro.
- `docs/specs/DISAMBIGUATION.md`: documents identity-slot resolution and the
  effective-scope configuration rule.

**Measured:** GB/T's diagnostic corpus (not fidelity-gated) moved 147/203 to
176/203. The full repository gate passes: formatting, clippy, and 2,471 tests.
The two new stack-alignment regressions cover scoped date configuration and
candidate-note suppression; the PR now carries 15 new tests in total.

**Stack:**

1. #1171 — candidate-neutral date-slot foundation (this PR)
2. Draft `DATE_SUBSTITUTE.md` specification
3. schema and engine implementation; spec becomes Active
4. GB/T style migration to named presets

Closes csl26-huuz. The broader mechanism is tracked by csl26-qbmd.

### DIFF
--- a/crates/citum-engine/src/processor/disambiguation.rs
+++ b/crates/citum-engine/src/processor/disambiguation.rs
@@ -55,10 +55,10 @@ use citum_schema::locale::Locale;
 pub struct Disambiguator<'a> {
     bibliography: &'a Bibliography,
     config: &'a Config,
-    /// Effective bibliography config governing multilingual/locale sort-key
-    /// policy (`sorting.multilingual`, `sorting.locale`, `sort-as`,
-    /// preferred transliterations). Year-suffix ordering must use this —
-    /// not `config` — so it agrees with the rendered bibliography order.
+    /// Effective bibliography config governing bibliography-owned date-slot
+    /// grouping and multilingual/locale sort-key policy. Year-suffix grouping
+    /// and ordering must use this — not `config` — when the bibliography
+    /// supplies the corresponding template or sort behavior.
     sort_config: &'a Config,
     locale: &'a Locale,
     group_sort: Option<&'a GroupSort>,
@@ -150,9 +150,9 @@ struct CachedReferenceData {
 impl<'a> Disambiguator<'a> {
     /// Creates a disambiguator that uses the default title-based fallback order.
     ///
-    /// `sort_config` is the effective bibliography config (multilingual/locale
-    /// sort-key policy); pass the same config used to sort the final
-    /// bibliography so year-suffix order agrees with the rendered order.
+    /// `sort_config` is the effective bibliography config; pass the same
+    /// config used to render and sort the final bibliography so bibliography-
+    /// owned date grouping and year-suffix order agree with its output.
     #[must_use]
     pub fn new(
         bibliography: &'a Bibliography,
@@ -175,9 +175,9 @@ impl<'a> Disambiguator<'a> {
 
     /// Creates a disambiguator with an explicit per-group sort specification.
     ///
-    /// `sort_config` is the effective bibliography config (multilingual/locale
-    /// sort-key policy); pass the same config used to sort the final
-    /// bibliography so year-suffix order agrees with the rendered order.
+    /// `sort_config` is the effective bibliography config; pass the same
+    /// config used to render and sort the final bibliography so bibliography-
+    /// owned date grouping and year-suffix order agree with its output.
     #[must_use]
     pub fn with_group_sort(
         bibliography: &'a Bibliography,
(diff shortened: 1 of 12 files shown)


crates/citum-engine/src/processor/disambiguation.rs
```diff
@@ -1018,16 +1018,163 @@ impl<'a> Disambiguator<'a> {
...
+                // this is the terminal case for this slot, not a signal to
+                // keep looking at a fallback chain.
+                return String::new();
+            }
+        } else if let Some(value) =
+            crate::values::date::resolve_date_variable(&component.date, reference)
+                .filter(|value| !value.is_empty())
+        {
+            return value.to_string();
+        }
```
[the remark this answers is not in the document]
bdarcus: Fixed — thanks for catching this. `date_component_discriminant` now calls a new `values::date::formatted_date_text` helper (form-restricted `format_date_range` + `apply_date_markers`, the same pipeline `TemplateDate::values` uses) instead of `DateValue`'s raw `Display`, for both the primary-variable branch and the fallback-loop `Date` candidate branch. Added a regression test (`given_a_copyright_fallback_date_with_day_precision_when_the_form_is_year_then_the_discriminant_is_the_bare_year`) proving two `copyright` dates with different day precision now collide under `form: year` instead of splitting. Re-verified: GB/T diagnostic corpus unchanged at 176/203, zero regressions across the 35-style exemplar corpus, all 2467 tests pass.

crates/citum-engine/src/processor/rendering/tests.rs
```diff
@@ -1239,3 +1239,138 @@ fn sentence_initial_group_still_capitalizes_leading_contributor_role_prose() {
...
+            }),
+            TemplateComponent::Title(TemplateTitle {
+                title: TitleType::Primary,
+                ..Default::default()
+            }),
+        ])
+    }
+
+    #[test]
+    fn given_two_undated_entries_with_an_accessed_fallback_when_they_collide_then_the_letter_lands_inside_the_wrap()
```
bdarcus: unit tests DO NOT USE BDD-STYLE naming! If this is a multi-parameter behavioral test, use `rstest.`

crates/citum-engine/src/processor/rendering/tests.rs
```diff
@@ -1239,3 +1239,138 @@ fn sentence_initial_group_still_capitalizes_leading_contributor_role_prose() {
...
+                String::new(),
+                "Anon. [2019b]. Title second".to_string(),
+            ],
+            "the disambiguating letter must sit inside the accessed date's own \
+             bracket wrap, not appended after it"
+        );
+    }
+
+    #[test]
+    fn given_two_undated_entries_with_no_fallback_content_when_they_collide_then_the_letter_renders_standalone()
```
bdarcus: unit tests DO NOT USE BDD-STYLE naming! If this is a multi-parameter behavioral test, use `rstest.`

crates/citum-engine/src/processor/disambiguation.rs
```diff
@@ -2336,4 +2529,278 @@ mod tests {
...
+        // (`fallback: [message: term.no-date]`) paths render the identical
+        // locale term via TemplateDate::values, so they must produce the
+        // identical collision-key discriminant — otherwise a style that
+        // mixes the two forms across type-variants would split undated
+        // references that render identical text.
+        assert_eq!(discriminant, "term:term.no-date:");
+    }
+
+    #[test]
+    fn given_an_empty_explicit_fallback_list_when_nothing_can_resolve_then_the_discriminant_is_empty()
```
bdarcus: unit tests DO NOT USE BDD-STYLE naming! If this is a multi-parameter behavioral test, use `rstest.`

crates/citum-engine/src/processor/disambiguation.rs
```diff
@@ -2336,4 +2529,278 @@ mod tests {
...
+        let locale = Locale::en_us();
+
+        let discriminant =
+            Disambiguator::date_component_discriminant(&component, &reference, &locale, None);
+
+        assert_eq!(discriminant, "1995");
+    }
+
+    #[test]
+    fn given_undated_references_with_differing_access_dates_when_grouped_then_presence_not_value_determines_collision()
```
bdarcus: unit tests DO NOT USE BDD-STYLE naming! If this is a multi-parameter behavioral test, use `rstest.`

crates/citum-schema-style/embedded/styles/gb-t-7714-2025-author-date.yaml
```diff
@@ -387,9 +387,13 @@ bibliography:
...
       suffix: '. '
-      fallback:
-      - message: term.no-date
-        form: short
+      # Unlike every other type-variant, upstream's `date-intext` macro never
+      # reaches the outer "no date" term for article-journal/article-magazine
+      # without volume/issue — its inner choose only ever falls through to a
+      # plain `issued` (or `available-date`) branch. An empty `fallback:`
+      # matches that: the date slot renders nothing rather than the locale's
+      # no-date term. See csl26-huuz.
```
bdarcus: Is this behavior justified? What is the implication behaviorally?

bdarcus: Implication: an undated `article-journal`/`article-magazine` reference renders **no date text at all** — not 无日期/n.d. like every other type-variant, genuinely blank — and if disambiguation still needs a letter, it renders standalone, alone in the date position. The comment now says that first, before the mechanism.

Justification: re-verified directly against the shipped upstream CSL's `date-intext` macro. Its `type="article-journal article-magazine"` branch is fully self-contained — none of its three inner sub-cases ever fall through to the outer no-date-term branch, because CSL/CSL-M `<choose>` selects one top-level branch and never continues past it. Whether GB/T *the standard* (not just this implementation) actually intends blank-vs-n.d. is a separate, open question — cited as csl26-9qat rather than resolved here; this is a style-data decision pinned to the current oracle, revisitable without an engine change.

On the `fallback: []` mechanism itself: I initially replaced it with a `suppress-no-date-term` flag, then reverted that per further discussion — it would just be one more thing for a planned `date-substitute` options mechanism (mirroring `author-substitute`, csl26-qbmd) to deprecate. That mechanism is being designed together with this PR's discriminant logic as a stacked follow-up, not bolted on independently. Added a test (`empty_fallback_list_leaves_the_date_position_blank_with_or_without_disambiguation`) covering both the no-collision and needs-disambiguation cases.

crates/citum-engine/src/processor/rendering/tests.rs
bdarcus: Fair — this was a plain single-scenario `#[test]`, not `#[rstest]`. Renamed to `accessed_fallback_letter_lands_inside_bracket_wrap`.

crates/citum-engine/src/processor/rendering/tests.rs
bdarcus: Same — renamed to `empty_fallback_chain_still_renders_letter_standalone`.

crates/citum-engine/src/processor/disambiguation.rs
bdarcus: Renamed to `empty_explicit_fallback_list_yields_empty_discriminant`.

crates/citum-engine/src/processor/disambiguation.rs
bdarcus: Renamed to `differing_access_dates_still_collide_via_shared_disambiguator`.

PR #1172, opened 2026-08-12
docs(spec): define date substitution

bdarcus:
Layer 2 of the date-substitute stack. This PR defines the public contract only; schema and engine implementation follow in Layer 3.

## Contract

- Defines `standard`, `gb-t-7714-2025`, and `gb-t-7714-2025-author-date` presets.
- Makes `standard` the default when a preset value is required while preserving omission as semantically distinct.
- Replaces the verbose options shape with a flat, ordered `TypeSelector` map and an intentional matched-empty state.
- Specifies eager preset expansion, per-selector inheritance merges, and whole-list candidate replacement.
- Preserves candidate `suppress-note` and requires rendering and disambiguation to share the same resolved candidate source and effective scope configuration.
- Limits substitution to the first identity date component whose `suppress-disamb-suffix` is not true.
- Registers the Draft and records the four-layer stack in `csl26-qbmd`.

## Stack

1. #1171 — candidate-neutral date-slot foundation
2. #1172 — Draft specification (this PR)
3. Schema and engine implementation; spec becomes Active
4. GB/T style migration to the new presets

## Validation

- `git diff --check`
- Relative documentation links inspected
- Documentation lint: no new violations (two unrelated pre-existing info-level findings)
- Bean hygiene passed (unrelated soft-stale advisories only)
- Commit-message hook passed
- Repository pre-push hook passed, including the full Rust quality gate
- Parent PR #1171: 16 successful checks, 1 skipped, 0 failures

### DIFF
--- a/docs/specs/DATE_SUBSTITUTE.md
+++ b/docs/specs/DATE_SUBSTITUTE.md
@@ -1,0 +1,357 @@
+# Date Substitute Specification
+
+**Status:** Draft
+**Version:** 1.0
+**Date:** 2026-08-12
+**Supersedes:** none
+**Related:** csl26-qbmd, PR #1171, [`DISAMBIGUATION.md`](./DISAMBIGUATION.md), [`PRIMARY_CONTRIBUTOR_SUBSTITUTION.md`](./PRIMARY_CONTRIBUTOR_SUBSTITUTION.md)
+
+## Purpose
+
+Define an options-level policy for the fallback candidates used when a
+reference's identity date is missing. The policy removes repeated inline
+`TemplateDate.fallback` chains, gives style families reusable named presets,
+and guarantees that rendering and disambiguation resolve the same candidate
+source under the same effective citation or bibliography configuration.
+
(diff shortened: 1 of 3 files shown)


docs/specs/DATE_SUBSTITUTE.md
```diff
@@ -0,0 +1,351 @@
...
+
+Define an options-level policy for the fallback candidates used when a
+reference's identity date is missing. The policy removes repeated inline
+`TemplateDate.fallback` chains, gives style families reusable named presets,
+and guarantees that rendering and disambiguation resolve the same candidate
+source under the same effective citation or bibliography configuration.
+
+A survey of the 32 embedded style YAML files found 64 `fallback:` keys, all in
+three GB/T files: 36 in `gb-t-7714-2025-base.yaml`, 25 in
+`gb-t-7714-2025-author-date.yaml`, and 3 in `gb-t-7714-2025-note.yaml`. Not all
```
bdarcus: What other uses of `fallback` are there? Do we need to consider adding support for others?

PR #1177, opened 2026-08-13
fix(styles): numeric bibliography marker gap

bdarcus:
## Summary
Three numeric styles (`springer-basic-brackets`, `springer-vancouver-brackets`, `taylor-and-francis-national-library-of-medicine`) declared `label-mode: numeric` without `label-wrap`/`label-separator`, so bibliography markers rendered unwrapped and flush (`1Kuhn TS...`). Values added per shipped CSL source. springer-vancouver also had an incorrect sentence-case title preset, fixed separately.

Exact-parity: 20→53/67, 11→40/67, 0→13/67. No regressions elsewhere (full-portfolio diff). Baseline regenerated; `csl26-l5oh` closed.

## Test plan
- [x] `cargo nextest run` (2501/2501)
- [x] `just check-core-quality` green
- [x] `node scripts/report-core.js --all-features` before/after, zero drift outside target styles


.beans/archive/csl26-l5oh--investigate-springer-vancouver-brackets-exact-pari.md
bdarcus, line 1: So this bean is more text than the actual fix. How is this useful? Rewrite to be no more than 100 words.

bdarcus: Trimmed to 62 words (72c4a119) — kept root cause, results, follow-up pointer; cut the investigation narrative.

PR #1206, opened 2026-08-18
docs(spec): make same-author collapse opt-in

bdarcus:
## Summary

**Docs-only spec, no code.** Adjudicates `csl26-ecfn` and, as a consequence,
resolves `csl26-m11m`.

- `citation.collapse` already exists and is genuinely opt-in — but it has
  exactly one variant (`citation-number`). Same-author collapse (author name
  once, years merged) is a *different* mechanism that's always-on in the
  engine and never consults this field.
- Corpus measurement: of 2,844 independent legacy CSL styles, **763 declare no
  `collapse` attribute at all** and Citum collapses them anyway — 44 numeric
  (invisible there), **361 note-regime** (the `csl26-m11m` bug). Migrate also
  silently discards `year-suffix`/`year-suffix-ranged` intent for 246 more
  styles today.
- citeproc-js ground truth on `chicago-notes-bibliography.csl` confirms notes
  never collapse; two adjacent Citum outputs already match it byte-for-byte,
  proving only the collapse *gate* is wrong, not the note rendering path.

**Decision:** extend the existing field with a `same-author` variant
(two-axis: *what* collapses vs. *how far*, mirroring the existing `Processing`
enum shape), default off. Note styles need **zero** note-specific code — they
simply never declare it. Full rationale, corpus tables, and three rejected
alternatives (regime-hack, boolean, CSL's flat enum) are in the spec.

## What's next

Implementation is a separate, stacked PR (schema change — this repo requires
the spec reviewed first). It adds the `Year`/`SameAuthor` schema variant,
gates the engine, fixes migrate's lossy mapping, and declares `collapse` on
the 5 embedded styles that need it to hold their current output steady.

Also surfaced, not filed: an independent locator-punctuation defect
(`4 (2019), 257` vs oracle `4 (2019): 257`) noticed while reproducing
`csl26-m11m` — reproduces on single-item citations, unrelated to collapse,
deliberately out of scope here.

## Test plan

- [x] `.md`-only change — no build gate required (repo policy)
- [x] `./scripts/validate-frontmatter.sh --copilot-strict` — clean except one
      pre-existing unrelated failure in a global (non-repo) skill file
- [ ] `gh pr checks` — docs-only CI


docs/specs/SAME_AUTHOR_COLLAPSE.md
```diff
@@ -0,0 +1,605 @@
...
+| `year-suffix-ranged` | 14 | 0 | 0 |
+
+`citation-number` collapse is overwhelmingly numeric, and the year forms are
+overwhelmingly author-date — but the lock is not 1:1 with `RegimeFamily`, and
+two regime families license *both* values:
+
+- **Label.** `din-1505-2.csl` declares `collapse="year"`;
+  `american-mathematical-society-label.csl` declares
+  `collapse="citation-number"`. Both are `Processing::Label` (the family
+  `has_citation_label` routes to before any other check runs).
```
bdarcus: This, per se, isn't relevant though. The semantics of this appear to me to be completely unclear. What is the meaning, and intended output, of a label style that specifies collapsing content that is not in the citation?

If there is a case here, you should demonstrate.

bdarcus: Checked. Both examples were wrong once I looked at the actual citation content instead of the attribute string:

- din-1505-2.csl is class="in-text" and its <citation> renders author-short + cite-year — an ordinary author-date citation. It's not Processing::Label at all (never renders variable="citation-label"); has_citation_label wouldn't route it there.
- american-mathematical-society-label.csl is genuinely Label-regime, but collapse="citation-number" is verified inert on it: stripped the attribute and re-rendered the same three-item cluster through citeproc-js — byte-identical output either way, `[Garc19, Garc19, Fort00]`. citeproc-js's numeric-range collapsing needs the template to render the number itself; a label-only template gives it nothing to collapse.

So there's no case, and Label is now excluded from both values in the regime-coherence rule. Rewrote the "Regime lock" section and §6 to drop the Label claim and state this directly, pushed in 52481a17.

PR #1231, opened 2026-08-24
docs(spec): bibliography substitute title quoting

bdarcus:
## Summary

**Docs-only spec — no engine/schema change in this PR.** Evaluates bean
`csl26-0u0f`: `chicago-author-date-18th` never quotes a bibliography-context
substituted title (author-less entries fall through to the title), diverging
from the citeproc-js oracle on ~16 rows. Per the bean's own recommendation and
this repo's schema-change policy, this PR is the reviewed spec; implementation
follows as a stacked PR.

- **New:** `docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md` — the
  bibliography-context companion to `SUBSTITUTED_VALUE_FORMATTING.md`, which
  is explicitly citation-scoped only.
- Re-measured the residual at this branch's base commit: **40 sole-cause / 18
  quote-only-clean rows** (down from the bean's 41/19 — `cd372f76`/`8cc47b8b`
  already fixed part of the same defect class). 17 are real after excluding
  one unrelated quote-character-normalization false positive; 16 are in scope
  after excluding one citation-context row.
- **Why the obvious category-level fix doesn't work:** all 16 rows resolve to
  Chicago's `component` title category, but that category isn't internally
  uniform — `map`/`graphic`/`classic`/`hearing` never quote via a type-template
  override. Traced `Rendering::merge`/`TitleRendering::merge` semantics to
  confirm a naive `titles.component.quote: true` fix would regress those four
  types, undoing this branch's own parent commits.
- Checked `taylor-and-francis-chicago-author-date-core` independently: same
  current residual, and its narrower `titles.type-mapping` (missing
  `manuscript`) is a concrete inheritance hazard for any type-keyed fix.
- Confirmed `apa-7th`/`elsevier-harvard` have **zero** bibliography-context
  defects of this kind today — no regression risk from whatever mechanism
  ships.
- **Recommendation:** an explicit per-type quote list on the bibliography
  substitute config, scoped to `chicago-author-date-18th` + its T&F
  descendant together. Two other mechanisms considered and rejected in the
  spec's Design section.

## Files changed

- `docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md` (new, Draft v1.0)
- `docs/specs/README.md` — registered under Text & Rendering
- `docs/specs/SUBSTITUTED_VALUE_FORMATTING.md` — v1.4, pointer-only (Related +
  Scope note to the new companion doc; no change to its own citation-scoped
  content)
- `docs/adjudication/DIVERGENCE_REGISTER.md` — div-011 note + changelog entry;
  contract unchanged
- Bean `csl26-0u0f` — progress note, stays `in-progress`

## Test plan

- [x] Docs-only — Rust pre-commit gate does not apply
- [x] Every number in the spec is reproducible from a quoted command
      (`report-core.js` + `analyze-parity-residuals.js` against
      `chicago-author-date-18th`, `taylor-and-francis-chicago-author-date`,
      `apa-7th`, `elsevier-harvard`)
- [x] Relative doc links resolve
- [x] CI green

## Next step

Once this spec is reviewed: implement mechanism (ii) as a stacked PR —
`chicago-author-date-18th` + `taylor-and-francis-chicago-author-date-core` in
one commit, `just schema-gen`, `report-core.js` before/after for all four
styles measured here, div-011 test block extended. Bean `csl26-0u0f` completes
in that PR, not this one.


docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md
```diff
@@ -0,0 +1,385 @@
...
+
+Drop the `RenderContext::Citation` half of the gate and let
+`resolve_category_quote` run in bibliography context too, reusing the
+existing `Substitute.title_quote` field (already resolved per-context via
+`options.config.effective_substitute()`,
+`crates/citum-engine/src/values/contributor/mod.rs:352` —
+`bibliography.options.substitute` already carries its own `candidates` and
+`contributor-role-form`, proving context-scoping needs no new schema field).
+
+**Rejected as insufficient on its own.** The reference-type partition above
```
bdarcus: Is "rejected" the right word here?

docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md
```diff
@@ -0,0 +1,385 @@
...
+Mitigated by requiring both to be edited together in the same commit
+(analogous to the "parent flip and required descendant compensation belong
+in the same commit" rule `SUBSTITUTED_VALUE_FORMATTING.md`'s Implementation
+Notes already sets), but the duplication itself is a standing maintenance
+cost this mechanism accepts rather than removes.
+
+### (iii) Derive substituted-title quoting from the reference type's own resolved template
+
+Instead of asking the category, ask the type's actual `title: primary` node
+in its resolved bibliography template whether it quotes — matching CSL
```
bdarcus: Why are you only focused on quoting? What about other common title formatting conventions like italics?

docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md
bdarcus: Reworded — v1.1 now reserves "rejected" for alternatives that are actually wrong (`titles.component.quote: true`, dropping the candidate, an unconditional global flip). The per-type quote list from v1.0 is now described as "superseded" / "insufficient once the emphasis-gap is in scope" rather than rejected — see the Rejected Alternatives section and §4's "why a per-type list no longer works."

docs/specs/SUBSTITUTED_TITLE_BIBLIOGRAPHY_QUOTING.md
bdarcus: You were right — this was a real gap, not a design choice. Traced it: the bean's residual was sized with `analyze-parity-residuals.js`'s "B title quote boundary" label, which regex-matches quote characters in markup-stripped diffs. Italic markup never survives that stripping into a diff op, so the tool used to size the original bean was structurally blind to any non-quote formatting mismatch — v1.0 inherited that blind spot rather than choosing it.

Direct comparison of markup-preserving oracle vs. Citum output found a real emphasis-gap (map/hearing/software/song/speech should italicize, render plain) alongside the quote-gap, both from one root cause: the substitute path only resolves title-category config, never a type's own resolved template node. Rewrote the spec (now `SUBSTITUTED_TITLE_BIBLIOGRAPHY_FORMATTING.md`) with a full taxonomy up front and changed the mechanism recommendation accordingly — deriving formatting from the type's own template node instead of a per-axis config list, since a per-type quote list would need a parallel per-type emphasis list too. Also surfaced (and explicitly scoped out, since it's a different defect surface) a candidate-selection gap: some article-magazine/newspaper rows should promote container-title, not title.

PR #1235, opened 2026-08-26
docs(spec): range/collapse config audit + spec

bdarcus:
## Summary
- Docs-only. Adds `docs/architecture/audits/2026-08-26_RANGE_COLLAPSE_CONFIG.md`: a full inventory of the 12+ config surfaces governing number-range collapsing (page ranges, date ranges, citation-number ranges, compound sub-labels), with file:line evidence for 7 concrete incoherences — most notably that `chi-chapter-locator` renders `112–118` instead of citeproc-js's `112–18` because `LocatorConfig::range_format` is a non-optional, hardcoded-`Expanded` field that shadows the style's `page-range-format: chicago16`
- Adds `docs/specs/RANGE_COLLAPSE_MODEL.md` (Status: Draft) — 5 design questions with candidate answers and rejected alternatives, raised for review before any implementation, per the project's schema-changes-need-a-docs-PR-first convention
- Files 3 beans under new epic `csl26-awlo`: `csl26-zepc` (this PR, completed), `csl26-rgys` (9 styles missing `collapse: citation-number`, separate defect), `csl26-7yxy` (Chicago styles hardcoding English prose/terms in `options.messages`, independent of the range work)

## Test plan
- [x] `./scripts/validate-frontmatter.sh --copilot-strict` — passes (one unrelated pre-existing failure outside this repo)
- Docs-only PR — skips build/test gate per CLAUDE.md

### DIFF
--- a/docs/specs/RANGE_COLLAPSE_MODEL.md
+++ b/docs/specs/RANGE_COLLAPSE_MODEL.md
@@ -1,0 +1,248 @@
+# Range/Collapse Model Specification
+
+**Status:** Draft
+**Version:** 1.0
+**Date:** 2026-08-28
+**Supersedes:** (none — no prior spec covers this ground)
+**Related:** `docs/architecture/audits/2026-08-26_RANGE_COLLAPSE_CONFIG.md`
+(the evidence behind this spec — read that first), `docs/specs/EDTF_DATE_RANGE_FORMATTING.md`
+(dates keep independent policy from pages, already shipped — see Decision 1),
+`docs/adjudication/DIVERGENCE_REGISTER.md` (`div-017`, `div-009` — the
+existing mechanism for intentional citeproc-js divergence, used in Decision
+2 and Acceptance Criteria), `csl26-awlo` (epic), `csl26-cl4q` (open-ended
+ranges, e.g. `"12ff"` — must fit this model), `csl26-rgys` (9 styles missing
+citation-number abbreviation — implementation should wait for this spec),
+`docs/policies/LOCALIZATION_INTEGRITY.md`
+
+## Purpose
+
+Citum abbreviates number ranges (`112–118` → `112–18`) correctly in one
+place — the bibliography page count — and inconsistently everywhere else a
+range can appear: citation locators, publication-date ranges, citation-number
+lists, same-author suffix runs, and compound sub-entries.
+
+**One range-rendering model: contextual inheritance, explicit overrides, and
+semantic defaults.** Two separate questions this spec keeps apart:
+
+- **Mechanism** — every range-producing surface uses the same
+  configuration-resolution system, with an explicit override always
+  available and nothing hardcoded at the call site. Decided: yes,
+  unconditionally (Decision 1).
+- **Default value inheritance** — what a surface gets when it doesn't
+  override. Not automatically the same value everywhere: a citation number
+  and a page number are both "numbers that can range," but they aren't the
+  same *kind* of thing. Resolved by the semantic-class layer in Decision 1
+  and the locator default in Decision 2.
+
+A rendered range also has two independent formatting properties that this
+spec keeps distinct rather than folding into one knob: **endpoint
+abbreviation** (`112–118` → `112–18`) and the **delimiter** character
+joining the two numbers (`–` vs `-` vs a locale-specific form). They often
+travel together but are resolved through separate chains (appendix).
+
+## Scope
+
+Configuration resolution for closed numeric ranges — endpoint abbreviation
+and delimiter selection — for page variables, citation locators,
+citation-number sequences, compound sub-entry sequences, date ranges, and
+same-author suffix sequences.
+
+Out of scope: whether adjacent values collapse into a range at all
+(`citation.collapse` and friends — untouched), open-ended ranges like
+`"12ff"` (`csl26-cl4q`), and the distinct parsing/formatting logic date and
+letter endpoints require beyond what they can share with numeric endpoints.
+
+**Nothing here removes an override a style author already has.** Every
+example below is the actual YAML. Locators, citation-number lists,
+same-author suffixes, and compound entries currently have *no* field to
+set their own abbreviation behavior — this spec gives them the same
+override mechanism `mhra-notes.yaml` already uses for locators today
+(`options.locators.range-format: expanded`).
+
+## Decisions
+
+| # | Decision | Resolution |
+|---|---|---|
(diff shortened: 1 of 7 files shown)


docs/specs/RANGE_COLLAPSE_MODEL.md
```diff
@@ -0,0 +1,182 @@
...
+correctly in one place — the bibliography page count — and inconsistently
+everywhere else a range can appear: citation locators, publication-date
+ranges, citation-number lists, same-author suffix runs, and compound
+sub-entries. This spec proposes one shared rule that all of those follow, and
+asks for a decision on five places where "one shared rule" still needs a
+judgment call.
+
+Nothing here is implemented yet. This is the decision to make before PR 2 of
+`csl26-awlo` writes any code.
+
```
bdarcus:
```suggestion
```

this doesn't belong here

docs/specs/RANGE_COLLAPSE_MODEL.md
```diff
@@ -0,0 +1,182 @@
...
+`pattern.page-range: "{$start}–{$end}"` is defined in five language files
+and read by nothing. It's effectively dead weight in the locale files with no
+current effect either way.
+
+- **Recommended: delete it.** The mechanism Citum actually uses for range
+  delimiters (a plain configured string) already works and is what every
+  other consumer would move to under decision 2. Keeping an unused,
+  differently-shaped mechanism around risks a future author reaching for it
+  and finding it does nothing.
```
bdarcus:
```suggestion
```

Does this belong here?

docs/specs/RANGE_COLLAPSE_MODEL.md
```diff
@@ -0,0 +1,182 @@
...
+everywhere else a range can appear: citation locators, publication-date
+ranges, citation-number lists, same-author suffix runs, and compound
+sub-entries. This spec proposes one shared rule that all of those follow, and
+asks for a decision on five places where "one shared rule" still needs a
+judgment call.
+
+Nothing here is implemented yet. This is the decision to make before PR 2 of
+`csl26-awlo` writes any code.
+
+## Decisions needed
```
bdarcus: you're completely missing the author angle, which is how this is all configured in the YAML styles. That was the focus of my concern, and you've ignored it.

docs/specs/RANGE_COLLAPSE_MODEL.md
bdarcus: Cut — that was process meta-commentary restating what `Status: Draft` already says in the frontmatter.

docs/specs/RANGE_COLLAPSE_MODEL.md
bdarcus: Checked: pattern.page-range is in fact a range-formatting artifact (a dead start–end template, not off-topic), so I kept the finding but demoted it out of the headline decision table — it's a minor dead-code cleanup, not a design decision on par with the other three. Now a one-line appendix note.

docs/specs/RANGE_COLLAPSE_MODEL.md
bdarcus: Rewrote the whole 'Decisions needed' section around this. Every decision now shows the actual current-vs-proposed YAML (chicago-author-date-18th's page-range-format/locators, mhra-notes' explicit locators.range-format override, the per-kind locators.kinds override, the citation/bibliography scoped-options block) instead of abstract prose about Rust settings. Also corrected a framing bug you flagged separately: I'd implied sharing one abbreviation rule across surfaces meant trading away configurability. It doesn't — every surface that has an override today keeps it verbatim; this only adds an override to the surfaces (locators, citation-number lists, same-author suffixes, compound entries) that currently have none at all.

docs/specs/RANGE_COLLAPSE_MODEL.md
```diff
@@ -0,0 +1,426 @@
...
+not about forcing every surface to inherit identical defaults.
+
+## Decisions needed
+
+| # | Decision | What changes for a reader | Recommendation |
+|---|---|---|---|
+| 1 | Give every kind of range — locators, citation-number lists, same-author suffixes, compound entries, dates — the same configuration-resolution mechanism that page counts already have, organized by semantic class rather than one flat default | Locators start abbreviating correctly (fixes the reported bug); every other kind of range becomes explicitly configurable instead of using a dash and a rule baked into the code with no YAML knob at all | Yes — do this. It's strictly additive: no style loses an override it has today |
+| 2 | Should the style-level abbreviation default apply to *every* citation locator kind (page, chapter, paragraph, note, …), or only to page numbers? | Whether `chapter 112–118` also gets shortened to `chapter 112–18` by default, or only bare page locators do | Open — CMOS's elision rule isn't inherently page-specific, but I found no evidence either way for citation-locator conventions specifically. Domain call, see below |
```
bdarcus: I see no reason to assume this is page number specific. I think we probably need to assume the answer to this question is "yes", but that we need to design it in such a way that it can be expanded later to override on a locator-specific basis.

docs/specs/RANGE_COLLAPSE_MODEL.md
```diff
@@ -0,0 +1,426 @@
...
+  options:
+    page-range-format: minimal
+bibliography:
+  options:
+    page-range-format: chicago16
+```
+
+This is expressible today. The delimiter is not scope-specific:
+`CitationOptions` and `BibliographyOptions` have no
+`page-range-delimiter` field. Their conversion produces `None` for that field,
```
bdarcus: we likely need to add a `range-delimiter` (not `page-range-delimiter`)?

docs/specs/RANGE_COLLAPSE_MODEL.md
bdarcus: Adopted verbatim: style-wide default applies to every locator kind. The per-kind override (locators.kinds.<kind>.range-format, already in the schema) is how a style opts a specific kind back OUT, not how it opts one in — that's the expansion path you're asking for, and it already exists, just with the polarity flipped from what I had.

docs/specs/RANGE_COLLAPSE_MODEL.md
bdarcus: Ran the corpus check the spec itself said would settle this: parsed every style's YAML (17 set page-range-format) for a citation.options/bibliography.options copy distinct from the style-level value. Zero hits. So the resolution is remove the scoped format entirely rather than add a sibling delimiter to it — nothing left for a scoped delimiter to pair with. Your range-delimiter naming still lands, just at the top level: page-range-format/page-range-delimiter -> range-format/range-delimiter, matching the generic name dates.range-format already uses in the shipped EDTF spec.

